PR #10, opened 2024-10-11
chore: Update build system from pip to uv

mortenoh:
Updates the build/dep system from `pip` to `uv`. All potentially unused files are left in `_unused`, we can clean those up later.

Lot's of very specific versions had to be specificed, we should look into upgrading some of those and also clean up unused deps.

We probably need to update all `Dockerfile` also as they are using `pip`.


knutdrand (approved):
Makefile will be used for testing, and some docs changes might break docs deployment, but we will see when docs gets pushed

PR #17, opened 2024-11-07
Fix-synch for TOGO

knutdrand:


chap_core/google_earth_engine/gee_era5.py
```diff
@@ -173,7 +173,7 @@ def _initialize_client(self):
 
         try:
             credentials = ee.ServiceAccountCredentials(account, key_data=private_key)
-            ee.Authenticate()
+            #ee.Authenticate()
```
hermantretteteig, line 176: do we not need to authenticate?

knutdrand: No, I don't think so. Ivar added it a couple days ago to make it work on gh actions. But it opens a browser when I use i

PR #30, opened 2025-02-04
Geometry simplify

karimbahgat:
Initial try at a standalone geometry simplify function that takes a Polygons object as input and returns a new Polygons that preserves topology, including a simple test. Should be tested on more and heavier geojsons, and several things that might need changing like the approach to testing, logging, and rendering, but making the PR now so we can start the review process. Before accepting the PR, we still need to figure out where the function should be called before sending simplified geometries to earth engine.


chap_core/geometry.py
```diff
@@ -156,10 +156,32 @@ class Polygons:
...
+    def __geo_interface__(self):
+        return self.to_geojson()
+
     @property
     def data(self) -> FeatureCollectionModel:
         return self._polygons
 
+    @property
+    def bbox(self):
+        from .geoutils import feature_bbox
```
karimbahgat, line 179: Had to do this to avoid circular import with geoutils.py, but not sure if this is in line with how you do this for the rest of chap-core / if you have any other suggestions?

chap_core/geoutils.py
```diff
@@ -0,0 +1,112 @@
...
+        xs, ys = zip(*exterior)
+        bbox = [min(xs),min(ys),max(xs),max(ys)]
+    elif geotype == "MultiPolygon":
+        xs = [x for poly in coords for x,y in poly[0]]
+        ys = [y for poly in coords for x,y in poly[0]]
+        bbox = [min(xs),min(ys),max(xs),max(ys)]
+    return bbox
+
+def render(polygons : Polygons):
+    import geopandas as gpd
```
karimbahgat, line 56: These imports are only needed for this functions so keeping them inside. But not sure if there are other conventions you use elsewhere in chap-core.

chap_core/geoutils.py
```diff
@@ -0,0 +1,112 @@
...
+    import matplotlib.pyplot as plt
+    from PIL import Image
+
+    df = gpd.GeoDataFrame.from_features(polygons.__geo_interface__)
+    fig, ax = plt.subplots(dpi=300)
+    df.plot(ax=ax)
```
karimbahgat, line 62: No styling yet, just simply showing the geometries

chap_core/geoutils.py
```diff
@@ -0,0 +1,112 @@
...
+#     # Convert image to NumPy array and display using Matplotlib (works in WSL)
+#     img_array = np.array(img)
+#     plt.imshow(img_array)
+#     plt.axis("off")  # Hide axes
+#     plt.show()
+
+def toposimplify(polygons : Polygons, threshold=None):
+    '''Simplifies a Polygons object while preserving topology between adjacent polygons.
+    
+    Args:
```
karimbahgat, line 88: Is there an agreed upon docstring style yet?

ivargr: We have been using the numpy docstring so far (https://numpydoc.readthedocs.io/en/latest/format.html), should probably be added to the contribution documentation:)

chap_core/geoutils.py
```diff
@@ -0,0 +1,112 @@
...
+        # calc as 1 thousandth of longest width or height of all polygons
+        frac = 0.001
+        xmin,ymin,xmax,ymax = polygons.bbox
+        w,h = xmax-xmin, ymax-ymin
+        longest = max(w,h)
+        threshold = longest * frac
+
+    # generate topology and simplify
+    kwargs = {'toposimplify':threshold}
+    topo = tp.Topology(polygons.__geo_interface__, **kwargs)
```
karimbahgat, line 106: This is where the topology is created and simplified. Right now only sets the `toposimplify` parameter, which sets the distance threshold used for simplifying. Other parameters that are also relevant and might need to be specified in the future are `prequantize`, `presimplify`, and `topoquantize`, see https://mattijn.github.io/topojson/example/settings-tuning.html#prevent_oversimplify.

tests/test_polygons.py
```diff
@@ -1,6 +1,15 @@
 import pytest
 import tempfile
+import json
+import pathlib
+import logging
 from chap_core.geometry import Polygons
+from chap_core.geoutils import toposimplify, render
+
+
+TEST_PLOTS_FOLDER = pathlib.Path(__file__).parent / '__test_plots__'
```
karimbahgat, line 10: For now putting test renders inside a temporary ignored output folder, but maybe there is already another approach to this used elsewhere?

tests/test_polygons.py
```diff
@@ -13,6 +22,42 @@ def test_to_from_geojson_file(data_path):
         assert polygons2 == polygons
 
 
+def test_geometry_simplify(data_path):
+    # load data
+    print('loading')
```
karimbahgat, line 27: Misc print statements for local testing but not crucial. Maybe there is already another approach for logging inside tests, or better to avoid overall?

tests/test_polygons.py
```diff
@@ -13,6 +22,42 @@ def test_to_from_geojson_file(data_path):
...
+    print(f'total geojson byte size reduced from {size1/1000}kb to {size2/1000}kb')
+    assert size2 < size1
+
+
 def test_laos_polygons(data_path):
     polygons = Polygons.from_file(data_path / "small_laos_data_with_polygons.geojson").data
     polygons = Polygons.from_file(data_path / "small_laos_data_with_polygons.geojson", id_property='district').data
+
+
+if __name__ == '__main__':
```
karimbahgat, line 60: Did this just for local debugging.

chap_core/geoutils.py
```diff
@@ -0,0 +1,112 @@
...
+    Args:
+    - polygons: Polygons object
+    - threshold: coordinate distance threshold used to simplify/round coordinates. If None, distance threshold
+        will be auto calculated relative to the bbox of all polygons, specifically one-thousandth of longest of bbox width or height. 
+    '''
+    import topojson as tp
+
+    # auto calc threshold if not given
+    if not threshold:
+        # calc as 1 thousandth of longest width or height of all polygons
```
karimbahgat, line 97: Although it is possible to specify the distance threshold in the function, it's probably best to leave it at None which will calculate it dynamically as a fraction of the total area covered by the Polygons object. So for a large country the threshold will be higher, and for a small country the threshold will be lower. The only case where this might be problematic would be some country where the districts are so small and many compared to the country as a whole that the simplification can significantly distort the district shapes. Don't think this will ever be the case, but could be something to test in the future.

chap_core/geoutils.py
karimbahgat: Switched and added instruction to docs

chap_core/geoutils.py
karimbahgat: Moved this comment to be a code comment instead

tests/test_polygons.py
karimbahgat: Switched to path fixture inside an ignored folder

PR #49, opened 2025-05-15
feat: add endpoint for creating backtest with dataset

eirikhaugstulen:
Summary:
- This PR introduces one workflow to create a dataset and immediately start an evaluation. 
- It should follow the same workflow for predictions.


tests/integration/rest_api/test_db_endpoints.py
```diff
@@ -356,3 +356,47 @@ def test_failing_jobs_flow(celery_session_worker, dependency_overrides):
     response = client.get(f'/v1/jobs/{job_id}')
     assert response.status_code == 200
     assert response.json() == 'FAILURE'
+
+
+def test_backtest_with_data_flow(
```
eirikhaugstulen, line 361: Tried to make a test, but it seems to only SKIP in CI. Not sure why.

chap_core/rest_api_src/v1/routers/analytics.py
```diff
@@ -324,3 +332,54 @@ class DataSource(DBModel):
...
+        if feature_name == "disease_cases":
+            continue
+        for location, data in provided_data_processed.items():
+            isnan = np.isnan(getattr(data, feature_name))
+            if np.any(isnan):
+                isnan_ = [data.time_period[i] for i in np.flatnonzero(isnan)]
+                raise HTTPException(
+                    status_code=500,
+                    detail=f"Missing value in {feature_name} in location {location}. Time periods: {isnan_}",
+                )
```
eirikhaugstulen, line 365: I don't fully understand everything that is happening here, so please have an extra look.

tests/integration/rest_api/test_db_endpoints.py
knutdrand: These tests require a running redis server for now. Karim is working on making it a bit more flexible

knutdrand:
Integration test is failing, but maybe that is just instability? Do they fail locally?

chap_core/rest_api_src/v1/routers/analytics.py
```diff
@@ -324,3 +332,54 @@ class DataSource(DBModel):
...
+
+    for feature_name in feature_names:
+        if feature_name == "disease_cases":
+            continue
+        for location, data in provided_data_processed.items():
+            isnan = np.isnan(getattr(data, feature_name))
+            if np.any(isnan):
+                isnan_ = [data.time_period[i] for i in np.flatnonzero(isnan)]
+                raise HTTPException(
+                    status_code=500,
```
Birkbjo, line 363: I think this should be a 409 conflict.

eirikhaugstulen: @knutdrand Using 409 for missing / conflicting data is a protocol we follow in the DHIS2 core api, but just following the rest of the code in this PR. Let me know if you want me to refactor anything.

knutdrand: We can do if that is more common. Putting this into a separate function anyway

PR #50, opened 2025-05-19
feat: add endpoint for renaming backtest

eirikhaugstulen:
Summary:
- Add a new PATCH endpoint
- Functionality for renaming a backtest


knutdrand:
Looks good, if any more patch functionality is added, we should maybe extract the pathcing logic into a separate functoin

PR #51, opened 2025-05-20
feat: add endpoint for batch deleting backtests

eirikhaugstulen:
Summary:
- Added a new endpoint that takes in a list of backtestIds and deletes them if present in the DB


knutdrand (commented):
should maybe be a router.delete endpoint, not post?
and is there any reason to wrap the ids in a class, or could it just be list[int]

eirikhaugstulen:
Hey @knutdrand!

I agree, and that was the original plan, but apparently there is some discussion regarding wether [delete request discourage the use of request bodies](https://stackoverflow.com/questions/299628/is-an-entity-body-allowed-for-an-http-delete-request) or not. The `class` was because I got a 422 validation error, and making it a POST fixed everything. 

I have no opinions here, so let me know what you prefer!

karimbahgat:
My immediate thought is that making it a delete endpoint with ids in the body would look/feel neater in the api and more consistent. Also maybe naming the function delete_backtest_batch or delete_backtest*s*?
image

I do see in the link you mention that some servers/load balancers may clean out the body of delete requests, but since chap is a bit of a smaller project with known deployment scenarios maybe we feel confident enough that we are willing to take that risk for now? 

Out of curiosity what is the use-case for the batch delete from the frontend?

knutdrand:
I think maybe the cleanest is to have the ids as query params and do it as
a delete request. That would solve both issues I think

PR #54, opened 2025-05-22
Filter out missing data

abyot:
When creating a dataset it is often case that an orgunit comes with no geojson data and hence no climate info. During this scenario, the API simply throws a 500 and reject the post request.

A pragmatic approach to this is to filter out - ignore - that specfic orgunit but continue with the rest. It is also informative for the API response to indicate which orgunit and what information are missing. 

Below is a sample dataset create post request response

```
{
    "id": "1626e6a8-02a7-4eb1-84d5-bc887bc43015",
    "imported_count": 355,
    "rejected": [
        {
            "reason": "Missing value for the entire time period",
            "orgUnit": "pS6VQHJLxbH",
            "feature_name": "mean_temperature"
        },
        {
            "reason": "Missing value for the entire time period",
            "orgUnit": "pS6VQHJLxbH",
            "feature_name": "rainfall"
        }
    ]
}
```

### DIFF
--- a/chap_core/rest_api_src/data_models.py
+++ b/chap_core/rest_api_src/data_models.py
@@ -37,7 +37,6 @@ class FetchRequest(DBModel):
     feature_name: str
     data_source_name: str
 
-
 class DatasetMakeRequest(DataSetBase):
     geojson: FeatureCollectionModel
     provided_data: List[ObservationBase]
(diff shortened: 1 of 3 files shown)


knutdrand:
I think we need to figure out how to handle regions with partly missing values before we roll it out. Right now it looks like we do a ValidationError every time at least one value is missing, but the error message is that it is missing values for the whole period. Should we raise an error when a region is missing only one value for instance?

knutdrand:
I think we should maybe the list should be one entry per region we reject, with maybe a list for which features was missing, and a list for which features had missing values maybe. This is a bit up to frontend @eirikhaugstulen @Birkbjo what makes more sense.

PR #84, opened 2025-09-11
feat: add model update endpoint

eirikhaugstulen:
Summary:
- Introduced a new endpoint to update model names while preserving template prefixes.
- Added `configuration_name` attribute to the `ModelSpecRead` class.


knutdrand:
Should probably also be a test for this in test_db_endpoints.py or similar

PR #126, opened 2025-12-15
yaml integration

AndreGregu:
created two scripts: produce_plots_from_yaml which uses a hard coded yaml file as an example of how to generate plots from a yaml file, and yaml_plot.py which should be a CHAP-compatible version of the code. in addition i updated the metrics rmse, crps and mae to calculate the correct error values based on median instead of mean, and also fixed peak_diff to support both montly and weekly data.


ivargr:
> The new features from this PR (YAML plotting and peak_diff monthly/daily support) have been extracted into PR #127.

Does that mean that this PR can be closed, @knutdrand ?

PR #231, opened 2026-03-04
feat: helm2

tonsV2:
# Description

The goal of this PR is to introduce Helm charts so CHAP can be deployed on IM. Due to the nature of the CHAP application (master/worker architecture) and some limitations of IM I had to split the chart into multiple charts all gathered under a main umbrella chart for easy deployment outside of IM. This split isn't something I would normally recommend but it makes it a lot simpler to reference individual components within IM.

# Review

@mortenoh, @karimbahgat - If the Helm charts are just seen as gibberish feel free to ignore them. But a brief review of the rest of the files would be appreciated :pray: 

@radnov - Show no mercy! :wink: 

# Before merge

During development I needed Docker images and Helm charts to be release so I updated the pipelines to run on this branch

* https://github.com/dhis2-chap/chap-core/blob/47ad27df7f76f8a478f8d0ff76989f9293105cac/.github/workflows/release-chart.yaml#L7
* https://github.com/dhis2-chap/chap-core/blob/47ad27df7f76f8a478f8d0ff76989f9293105cac/.github/workflows/ci-build-docker-images.yml#L5

The above should be reverted before merging


.github/workflows/ci-build-docker-images.yml
```diff
@@ -2,7 +2,7 @@ name: Create and publish chap Docker image
 
 on:
   push:
-    branches: ["dev", "master"]
+    branches: ["dev", "master", "feat/helm2"]
```
radnov: Reminder to delete this before merging.

charts/chap/Chart.lock
radnov, line 1: The lock file seems to be stale. The `chap-db` and `chap-api` versions are older compared to the `Chart.yaml` file.

charts/chap-api/templates/secrets.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ include "chap-api.fullname" . }}
+  labels:
+    {{- include "chap-api.labels" . | nindent 4 }}
+type: Opaque
+data:
+  GOOGLE_SERVICE_ACCOUNT_EMAIL: {{ .Values.googleServiceAccount.email | b64enc | quote }}
+  GOOGLE_SERVICE_ACCOUNT_PRIVATE_KEY: {{ .Values.googleServiceAccount.privateKey | b64enc | quote }}
```
radnov: Even if the `email` is not set/empty the private key will be created. Not a big deal. but we could add 

```
{{- if .Values.googleServiceAccount.email }}
```

charts/chap-api/templates/register-job.yaml
```diff
@@ -0,0 +1,64 @@
...
+metadata:
+  name: {{ include "chap-api.fullname" . }}-register-{{ .Release.Revision }}
+  labels:
+    {{- include "chap-api.labels" . | nindent 4 }}
+spec:
+  ttlSecondsAfterFinished: 3600
+  backoffLimit: 0
+  template:
+    spec:
+      restartPolicy: OnFailure
```
radnov: Aren't the `backoffLimit` and `restartPolicy` contradictory?

The shell script already has retry logic (`MAX_ATTEMPTS=10`), but with `backoffLimit: 0` the Job will fail after one pod failure, yet `restartPolicy: OnFailure` will restart the container within the pod. 

Maybe we need `restartPolicy: Never` and `backoffLimit` with the actual number of pod-level retries, or keep `OnFailure` and raise `backoffLimit`. 🤔

charts/chap-api/templates/register-job.yaml
```diff
@@ -0,0 +1,64 @@
...
+                  -H "Content-Type: application/json" \
+                  -u "${username}:${password}" \
+                  --connect-timeout 10 \
+                  --max-time 30 \
+                  --data "$PAYLOAD" \
+                  "https://${hostname}/api/42/routes")
+                HTTP_CODE=$(printf '%s' "$RESPONSE" | tail -1)
+                BODY=$(printf '%s' "$RESPONSE" | head -n -1)
+                # TODO: Support other DHIS2 versions than 42
+                echo "POST https://${hostname}/api/42/routes -> HTTP $HTTP_CODE: $BODY"
```
radnov: Lets expose the DHIS2 version as a chart value, even though other versions besides 42 might not be supported now.

compose.yml
```diff
@@ -1,5 +1,6 @@
 services:
   chap:
+    image: ghcr.io/dhis2-chap/chap-core:latest
```
radnov: With both `image` and `build` set the former will take precedence when running `docker compose up` and will pull the image from ghcr.io, which might not be intended for local development. i.e. building a local image requires the explicit `--build` option, so we should probably document this.

charts/chap-api/values.yaml
```diff
@@ -0,0 +1,122 @@
...
+  port: 6379
+  password: ""
+
+googleServiceAccount:
+  email: ""
+  privateKey: ""
+
+dhis2:
+  enabled: true
+  hostname: ""
```
radnov, line 118: Should we have a default hostname, given that this is enabled by default?

charts/chap-db/templates/cluster.yaml
```diff
@@ -0,0 +1,42 @@
...
+    size: {{ .Values.postgresql.cnpg.cluster.storageSize }}
+  {{- if .Values.postgresql.cnpg.cluster.database }}
+  bootstrap:
+    initdb:
+      database: {{ .Values.postgresql.cnpg.cluster.database }}
+      owner: postgres
+      secret:
+        name: {{ include "chap-db.postgres.secretName" . }}
+  {{- end }}
+  enableSuperuserAccess: true
```
radnov: Should we expose this as a value?

.github/workflows/release-chart.yaml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - master
+      - feat/helm2
```
radnov: Reminder 🔔

radnov (changes requested):
Looks good! It's a lot of YAML to review, so I might have skipped over templating parts, but added a couple of comments.

charts/chap-api/values.yaml
tonsV2: I disabled it by default since we have no way of knowing the hostname

radnov (approved):

charts/chap-api/templates/deployment.yaml
```diff
@@ -54,90 +54,80 @@ spec:
...
             - name: POSTGRES_DB
               valueFrom:
                 configMapKeyRef:
-                  name: {{ include "chap.fullname" . }}
+                  name: {{ include "chap-api.fullname" . }}
                   key: POSTGRES_DB
             - name: CHAP_DATABASE_URL
               value: "postgresql://$(POSTGRES_USER):$(POSTGRES_PASSWORD)@$(POSTGRES_HOST):$(POSTGRES_PORT)/$(POSTGRES_DB)"
+            {{- if .Values.googleServiceAccount.email }}
+            - name: GOOGLE_SERVICE_ACCOUNT_EMAIL
```
mortenoh: The google keys can be fully removed now, chap-core no longer depends on or has code that uses GEE

charts/chap-api/values.yaml
```diff
@@ -0,0 +1,125 @@
...
+
+image:
+  repository: ghcr.io/dhis2-chap/chap-core
+  pullPolicy: IfNotPresent
+  tag: ""
+
+replicaCount: 1
+
+command:
+  - uvicorn
```
mortenoh: any good reason to switch to uvicorn instead of gunicorn (which uses the uvicorn worker), we used this approach as we have custom code in gunicorn.conf.py

chap_core/rest_api/app.py
```diff
@@ -25,9 +26,19 @@
     {"name": "Services", "description": "Service registry (v2)"},
 ]
 
+
+@asynccontextmanager
```
mortenoh: This should not be needed anymore that we are back to using gunicorn as the gunicorn.conf.py handles the database startup (we want this to be up and running before the worker threads start up). We only do 1 worker now, but we hope to fix this later on.

mortenoh (changes requested):

mortenoh (approved):

PR #262, opened 2026-03-30
Add explainability module with LIME

Rednael18:
This pull request introduces an isolated module for model agnostic prediction explanations using LIME.  Almost all additions in this PR are in the new chap_core/explainability/ folder. See userguide.md for an introduction to LIME and how to use the module. 

In addition to this folder, an endpoint has been added to the CLI (in chap_core/cli_endpoints/explain.py) called explain_lime.


knutdrand:
Nice work on this -- the module is impressively comprehensive and the userguide is really helpful for understanding the design. I know the explainability code itself is still maturing, so I focused my review on a few things that touch the rest of the codebase and would be good to address before merge:

1. **Mutating `_non_feature_names` from `model_spec.py`** (`chap_core/explainability/lime.py:368-369`):
   ```python
   _non_feature_names_plus_disease = _non_feature_names
   _non_feature_names_plus_disease.remove("disease_cases")
   ```
   This is an alias rather than a copy, so `.remove(...)` mutates the original set in `model_spec.py` in place at import time. Since `cli.py` now imports `explain` unconditionally, this happens on every CLI invocation. Easy fix:
   ```python
   _non_feature_names_plus_disease = _non_feature_names - {"disease_cases"}
   ```

2. **Eager CLI imports of the LIME stack** (`chap_core/cli.py`): registering `explain` at the top level means `stumpy`, `pyts`, `fastdtw`, `matplotlib`, etc. get imported on every `chap` invocation (`chap --help`, `chap evaluate`, ...), even when nobody is running `explain-lime`. Could we lazy-import inside the command function so the cost is only paid when the explain command is actually used?

3. **New mandatory dependencies** (`pyproject.toml`): `stumpy`, `pyts`, and `fastdtw` are added to the base dependencies, which means every install of `chap-core` (Docker images, REST API runtime, downstream consumers) pulls them in. Since the explainability module is opt-in and still experimental, would it make sense to put these under an optional `[explainability]` extra and import them lazily?

4. **`DataSet.__add__`** (`chap_core/spatio_temporal_data/temporal_dataclass.py:679-687`): this adds a `+` operator to a core class used throughout the codebase, but the implementation drops both `polygons` and `metadata` (the sibling `rename_location` and `iter_locations` methods in this same diff correctly preserve them). Since `produce_lime_dataset` could just construct a `DataSet(...)` directly, one option is to drop `__add__` from this PR and revisit it separately when there's a clear contract for merging polygons/metadata.

A couple of smaller things while I'm here:

5. `chap_core/explainability/perturb.py:17-19` calls `logging.basicConfig(...)` at module import time. This is a global side effect that can override chap's own logging config (set up via `initialize_logging`) -- modules generally shouldn't call `basicConfig`.

6. The `docker_helper_functions.py` change looks unrelated to explainability -- might be worth splitting into its own PR so it doesn't get lost in the diff.

7. `iter_locations` and `rename_location` are useful additions to `DataSet` -- would be nice to have a small unit test for them since they're now part of the public API.

Happy to chat about any of these if helpful. The cross-cutting items (1-4) are the ones I'd consider blocking; everything inside the explainability module itself is fine to refine later.

knutdrand:
Great,  can you run ruff on it (CI is failing on ruff now)

PR #264, opened 2026-04-06
Save model_info in Evaluation NetCDF metadata

bsmarkus:
Currently, the model card generator(not merged yet) can not access the model template model_info from the model template directly, and requires the run directory as an additional input. This PR stores serialized model_info in Evaluation NetCDF metadata during Evaluation.to_file( ), so model card generation can use only the evaluation file path.

### DIFF
--- a/chap_core/assessment/evaluation.py
+++ b/chap_core/assessment/evaluation.py
@@ -597,6 +597,7 @@ class Evaluation(EvaluationBase):
             model_name: Name of the model (optional)
             model_configuration: Model configuration dictionary (optional)
             model_version: Model version string (optional)
+            model_info: ModelTemplateConfigV2 object containing model metadata (optional)
         """
         flat_data = self.to_flat()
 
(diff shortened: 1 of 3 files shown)


knutdrand:
Docstring for `to_file` (`evaluation.py:592-596`) is missing the new `model_info` parameter in the Args section.

knutdrand:
A few related issues around how `model_info=None` is handled:

1. **Empty-string sentinel** (`evaluation.py:603`): When `model_info` is `None`, `""` gets written to `ds.attrs["model_info"]`. Consumers then have to do truthy checks, and `ModelTemplateConfigV2.model_validate_json("")` would raise. Cleaner to omit the attr entirely when `None` (don't add the key to `model_metadata`), so the round-trip is unambiguous.

2. **`model_info or None` is dead code** (`cli_endpoints/evaluate.py:432`): `estimator.model_information` is already typed `ModelTemplateConfigV2 | None`, so the `or None` is a no-op -- and would silently swallow any future falsy non-`None` value. Just pass `model_info=model_info`.

Suggested minimal fix:
```python
# evaluation.py
if model_info is not None:
    model_metadata["model_info"] = model_info.model_dump_json(exclude_none=True)
```
and only set the attr in `_flat_data_to_xarray` when the key is present.

bsmarkus:
Addressed feedback :)

PR #266, opened 2026-04-07
feat: add regional rmse distribution plot

bsmarkus:
- Adds a new regional RMSE distribution plot
  - Plots boxplots for each region in the evaluation, showing the error distribution.
  - Will be used in the model card generator to compare performance across regions
- Added unit test

### DIFF
--- a/chap_core/assessment/backtest_plots/__init__.py
+++ b/chap_core/assessment/backtest_plots/__init__.py
@@ -313,6 +313,7 @@ def _discover_plots():
         horizon_location_grid,
         metrics_dashboard,
         predicted_vs_actual_plot,
+        regional_rmse_distribution,
         sample_bias_plot,
     )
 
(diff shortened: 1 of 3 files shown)


knutdrand:
1. Typo in name: "Regional RMSE distributioion" (double "io") in the `@backtest_plot` decorator (`regional_rmse_distribution.py:25`).
2. Typo in test name: `test_regional_rmse_distribtuion_plot_directly` ("distribtuion") (`test_backtest_plot.py:114`).
3. Tooltip format `.75f` (`regional_rmse_distribution.py:68`): The mean RMSE tooltip uses `format=".75f"` which would show 75 decimal places. This is almost certainly a mistake -- should probably be `.2f` or `.4f`.
4. Duplicate title: Both `by_location.properties(title=...)` and the final `chart.properties(title=...)` set the title, with slightly different casing ("RMSE Distribution by Location" vs "RMSE Distribution By Location"). The second one will override the first.

bsmarkus:
Feedback addressed, let me know what you think now :)

PR #272, opened 2026-04-10
Hpo/eval

zlilu:


knutdrand:
* update dry_run logic: if dry_run mode= Normal
* have normal in enum for model type
* merge eval and eval_hpo
* export should use calculate
* don't export in objective
* try/except/raise from in objective

PR #274, opened 2026-04-11
feat: add regional mape distribution plot

bsmarkus:
Identical to RMSE regional distribution plot, but for MAPE instead.


knutdrand:
Should probably extract common functionality for rmse and make into a common base class in order to remove code duplication. Maybe worth creating these as Metric Plots instead?

PR #288, opened 2026-04-15
feat: Added make_plot method to Evaluation Class

bsmarkus:
Support generating metric plots from Evaluation instance with a new method make_plot.

Ref: https://github.com/dhis2-chap/chap-core/pull/274#issuecomment-4235979207


bsmarkus:
Add as a standalone function in evaluation_plot.py instead

PR #296, opened 2026-04-20
feat: add endpoint to download backtest metrics as CSV [CLIM-598]

eirikhaugstulen:
## Summary
- Adds `GET /v1/crud/backtests/{backtestId}/metrics/csv` returning a long-format CSV with per-location / per-time_period / per-horizon values for every applicable metric in `available_metrics`.
- Introduces `compute_all_detailed_metrics_from_backtest` in `chap_core.assessment.metrics`, which iterates the registered metrics, skips those that aren't applicable, and tolerates individual metric failures so one broken metric doesn't take down the whole export.

## Test plan
- [x] `uv run pytest tests/evaluation/test_metrics.py::test_compute_all_detailed_metrics_from_backtest`
- [x] `uv run pytest tests/integration/rest_api/test_from_seeded_engine.py::test_backtest_metrics_csv tests/integration/rest_api/test_from_seeded_engine.py::test_backtest_metrics_csv_missing`

Demo: 

https://github.com/user-attachments/assets/ed40c0ec-a914-43e6-8469-a69218bc88eb

### DIFF
--- a/chap_core/assessment/metrics/__init__.py
+++ b/chap_core/assessment/metrics/__init__.py
@@ -203,18 +203,15 @@ def compute_all_aggregated_metrics_from_backtest(backtest: BackTest) -> dict[str
     return results
 
 
-def compute_all_detailed_metrics_from_backtest(backtest: BackTest) -> pd.DataFrame:
+def compute_all_detailed_metrics(evaluation: Evaluation) -> pd.DataFrame:
     """
-    Compute all applicable metrics at detailed resolution for a backtest.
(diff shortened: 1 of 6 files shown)


knutdrand:
Maybe make the signature for the computation function accept an Evaluation object. (created from .from_backtest or init). Then the signature gets similar to the calculate_metrics functions below

knutdrand:
And maybe the endpoint path could be metric/ first, so that we can seamlessly expand this to accept multiple evaluations later?

PR #297, opened 2026-04-20
feat: add configured model info endpoint

eirikhaugstulen:
## Summary

- Adds `GET /v1/crud/configured-models/{configuredModelId}` returning a configured model together with its parent template, so the frontend can render the user-option schema (e.g. the `n_lags` dynamic list) alongside the stored values without stitching multiple list calls.
- Moves the display-name stitching from `SessionWrapper.get_configured_models` into a `ConfiguredModelDB.display_name` property so it can be reused by the new detail view.
- Relocates `ModelTemplateRead` next to the new `ConfiguredModelInfoRead` in `model_templates_and_config_tables`.

## Test plan

- [x] `uv run pytest tests/test_database.py::test_configured_model_display_name`
- [x] `uv run pytest tests/integration/rest_api/test_db_endpoints.py::test_get_configured_model_info` (skipped locally without celery worker; exercised in CI)
- [x] `uv run mypy` on changed files


knutdrand (commented):
This seems good to me. I think it steps a bit on the toes on a refactor PR from @ferenckata moving some data models, so maybe the merge will be cleaner if you revert moving date data model class Definitions. If this now replaces an endpoint, maybe also throw on a deprecated tag on the old one.  Also @mortenoh made an experimental endpoint decorator recently that we can use on endpoints under construction. Feel free to merge in the cleanest way possible. And in the future, new endpoints with experimental tags are good to merge if pr from me is blocking

PR #300, opened 2026-04-22
CLIM-592 moved pydantic models to data_models, linted tests and uv lock

ferenckata:
## Summary

CLIM-592 — consolidate REST API pydantic request/response models in one module.

- Move inline `BaseModel` / `DBModel` classes out of `chap_core/rest_api/v1/routers/{analytics,crud}.py` into `chap_core/rest_api/data_models.py`.
- Update all downstream callers (test conftests, `test_db_endpoints.py`, `scripts/anonymize_request.py`) to import the moved classes directly from `data_models` rather than relying on re-exports from the router modules.
- Drop three unused models picked up during the move (`EvaluationEntryRequest`, `MetaDataEntry`, `MetaData`) — none of them are referenced anywhere.
- Lint test files and refresh `uv.lock` (`fastdtw` entry).

No semantic changes — moved class definitions are unchanged.

## Test plan

- [x] `uv run pytest` — 522 unit tests pass
- [x] Integration test collection (160 tests collect cleanly)
- [x] `uv run ruff check` clean on changed files


ferenckata:
There is also a JobDescription in celery_tasks.py, and some more in common_routes.py and worker_functions.py too. Should these also move to data_models?

And there is a dataclass DataValue in worker_functions, maybe this should be pydantic instead?

-> following chat with Knut, the answer is no, these are different as they are not API endpoints (or something like that)

knutdrand:
**Downstream imports not updated**

Several call-sites still import the moved models from the old router modules, relying on the re-export as a side-effect:

- `tests/conftest.py:25-26` → `DatasetCreate`, `MakePredictionRequest` from `crud`/`analytics`
- `tests/integration/conftest.py:7` → `MakeBacktestWithDataRequest` from `analytics`
- `tests/integration/rest_api/test_db_endpoints.py:20-21` → `MakePredictionRequest`, `DatasetCreate`, `ModelConfigurationCreate`, `ModelTemplateRead`
- `scripts/anonymize_request.py:7` → `MakePredictionRequest`

These will work today (the names are re-exported when the router imports them), but the point of centralizing is to have one canonical import path. Worth updating these to `from chap_core.rest_api.data_models import ...` in this PR so the move is complete and future removals from the routers won't silently break callers.

ferenckata:
ah, good point. Should I merge Eirik's branch to this first?

mortenoh:
> ah, good point. Should I merge Eirik's branch to this first?

No, I don't think thats needed. Just fix the merge conflicts, and update as @knutdrand said, then I'll quickly review and we can merge it in.

mortenoh (approved):

PR #303, opened 2026-04-22
feat: Added function for making altair plots from evaluation objects

bsmarkus:
- Added make_plot_from_evaluation_object as a function
- Added unit test

ref comment:
https://github.com/dhis2-chap/chap-core/pull/288#issuecomment-4297168712


bsmarkus:
@knutdrand Ready for review, i noticed there is no button to request review so i notified you here instead.
image

PR #307, opened 2026-04-23
feat: Added generate-modelcard CLI command

bsmarkus:
- Added a new CLI endpoint for generating a model card:
  - Takes an evaluation NetCDF file as an input and generates a markdown report
  - Optionally takes a GeoJSON to plot regional aggregate RMSE and MAPE maps.
- Registered command and added unit test

### DIFF
--- a/chap_core/cli.py
+++ b/chap_core/cli.py
@@ -15,6 +15,7 @@ from chap_core.cli_endpoints import (
     evaluate,
     explain,
     forecast,
+    generate_modelcard,
     preference_learn,
     utils,
     validate,
@@ -36,6 +37,7 @@ forecast.register_commands(app)
 preference_learn.register_commands(app)
 utils.register_commands(app)
 validate.register_commands(app)
+generate_modelcard.register_commands(app)
 
 
 def main():
--- a/chap_core/cli_endpoints/generate_modelcard.py
+++ b/chap_core/cli_endpoints/generate_modelcard.py
@@ -15,6 +15,7 @@ from chap_core.assessment.metrics import (
     RMSEMetric,
     compute_all_aggregated_metrics_from_backtest,
 )
+from chap_core.database.model_templates_and_config_tables import ModelTemplateMetaData
 from chap_core.external.model_configuration import ModelTemplateConfigV2
 from chap_core.plotting.evaluation_plot import (
     MetricByTimePeriodV2Mean,
@@ -29,6 +30,13 @@ except ImportError:
     CHAP_VERSION = "unknown"
 
 MISSING = "More Information Needed"
+
+PLACEHOLDER_METADATA_VALUES = {
+    "display_name": ModelTemplateMetaData.model_fields["display_name"].default,
+    "description": ModelTemplateMetaData.model_fields["description"].default,
+    "author_note": ModelTemplateMetaData.model_fields["author_note"].default,
+    "author": ModelTemplateMetaData.model_fields["author"].default,
+}
 
 
 def is_url(path_or_url: str | Path) -> bool:
@@ -101,6 +109,20 @@ def _load_dataset_attrs(evaluation_path: Path) -> dict[str, Any]:
             "model_config": _parse_model_config(ds.attrs.get("model_configuration")),
             "historical_context_periods": int(ds.attrs.get("historical_context_periods", 0)),
         }
+
+
+def _normalize_metadata_value(value: str | None, field_name: str) -> str | None:
+    if not value:
+        return None
+
+    stripped_value = value.strip()
+    if not stripped_value:
+        return None
+
+    if stripped_value == PLACEHOLDER_METADATA_VALUES.get(field_name):
+        return None
+
+    return stripped_value
 
 
 def _save_evaluation_plots(evaluation: Evaluation, output_dir: Path, geojson_path: Path | None) -> None:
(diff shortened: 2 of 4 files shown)


bsmarkus:
@knutdrand Ready for review :)

knutdrand:
_This review was generated by Claude._

### Review: PR #307 — `feat: Added generate-modelcard CLI command`

**CI**: All green (ruff + build across matrix).

#### Overview
Adds a new `chap generate-modelcard` CLI command that reads an evaluation NetCDF, rebuilds a `Backtest`, computes aggregate metrics + plots, and writes a HuggingFace-style Markdown "model card" (plus PNG/HTML side artefacts). With an optional GeoJSON, it also renders regional RMSE/MAPE choropleths.

Registration follows the existing `register_commands` convention (`cli.py:18,40`), matching the other endpoints in `chap_core/cli_endpoints/`.

---

#### Correctness issues (worth fixing)

- **`ModelTemplateMetaData` has non-None string defaults** — `chap_core/database/model_templates_and_config_tables.py:22-31` assigns `display_name="No Display Name yet"`, `description="No Description yet"`, `author_note="No Author note yet"`, `author="Unknown Author"`. In `generate_modelcard.py`:
  - L229: `if (meta_data) and (meta_data.author_note):` will always be true when meta is present, so the card prints the literal string `"No Author note yet"`. Same for the `description` block at L241-242 and `author`/`display_name` fallbacks.
  - Either compare against the known placeholder strings, or treat those placeholders as "missing" up front. As it stands the placeholder text leaks into published model cards.
- **`author_assessed_status` fallback is dead** (L205-207). The field defaults to `AuthorAssessedStatus.red`, so the `else None` branch never fires and every generated card reads a red status even when the author didn't set one.
- **No validation that `user_option_values` / additional covariates came from the right config** — `_parse_model_config` silently returns `{}` on bad input, but you never warn the user if the NetCDF had no `model_configuration` attr. Consider logging a warning so silent "MISSING" outputs are traceable.
- **`try: from chap_core import __version__` fallback** (L28-31) — `chap_core` is the host package; this can never `ImportError` at runtime. Drop the try/except or read version via `importlib.metadata.version("chap-core")`.
- **`documentation_url` on `ModelTemplateMetaData` exists but is never rendered.** It'd be a natural fit for the "More information" section.

#### Test coverage (CLAUDE.md item #9)
- Only one test and it only checks that the `.md` file exists and contains one image reference. It does **not** exercise:
  - the GeoJSON branch (whole `if geojson_path:` block untested);
  - that plot artefacts (`eval_plot.png`, `detailedRMSE_plot.png`, map outputs…) actually get written;
  - the metadata-rendering branches (most of the logic).
- `content.__contains__(...)` — use `in` (idiomatic, matches the rest of the test suite).
- `output_file` is declared then re-built inline (`tmp_path / "modelcard.md"`), and `open(output_file)` should be `output_file.read_text()`. Minor.
- Test file lives in a new `tests/evaluation/generate_modelcard/` subdirectory without a `conftest.py`; it works because `backtest` is defined in `tests/evaluation/conftest.py` and pytest walks upward. Fine — just flag that the extra directory level isn't really buying anything (the neighbouring `test_evaluation_*.py` files live directly in `tests/evaluation/`).

#### Style / conventions
- File is 445 lines; ~200 of those are static markdown template string appends. The function `generate_modelcard` has grown close to 200 lines on its own. Consider extracting the template body into a small helper (`_render_markdown(context) -> str`) or a Jinja2 template — easier to test and unit-check.
- `_save_evaluation_plots` writes **14 files** (7 plots × {png, html}) always. PNG export needs `vl-convert-python`; if that's not installed the command will blow up with an Altair error deep in the stack. Either assert the dep up front or make PNG optional via a flag.
- Minor: `cast("dict[str, Any]", parsed)` uses a string-literal type argument; `cast(dict[str, Any], parsed)` is more common.
- `md.append(...)` pattern with lots of interleaved empty strings is hard to read; `md += [..., ..., ""]` or a joined multi-line string literal block would shrink this meaningfully.
- Per CLAUDE.md #7 (be concise): the whole HF template is copied verbatim including lots of `[optional]` empty headings (Glossary, Model Card Authors, Environmental Impact…). None of this is populated from the evaluation file. If the intent is to emit a stub for humans to fill in, fine — but worth confirming that the scope isn't bloated.

#### Performance / safety
- No check that `evaluation_path` / `geojson_path` exist; you'll get a bare `xr.open_dataset` / `Path.read_text` stack trace. A friendlier early error would help.
- `geojson["features"]` — `KeyError` if the user passes a GeoJSON that isn't a FeatureCollection. Consider validating or catching.
- `output_file.with_suffix(".md")` silently renames e.g. `card.txt` → `card.md`. Probably fine, but worth a one-line docstring note so users aren't surprised when `--output-file foo.html` produces `foo.md`.

#### Security
- No user-supplied strings are executed; markdown injection is possible via `meta_data.*` fields (someone could set `author_note` to raw HTML/`<script>`), but that's the same trust model as the DB today. Not a new risk.

---

#### Suggested asks before merge
1. Handle the default-placeholder strings in `ModelTemplateMetaData` so cards don't print "No Description yet"/"No Author note yet"/"Unknown Author".
2. Fix the always-red `author_assessed_status` fallback (or drop the fallback).
3. Add at least one test for the `geojson_path` branch and assert that a couple of plot files were produced.
4. Drop the spurious `try/except ImportError` for `__version__`.
5. Consider rendering `documentation_url`.

Otherwise the structure is clean, the command registers properly, and the CI signal is good — this is close to mergeable once the metadata-defaults issue is addressed.

bsmarkus:
@knutdrand Feedback addressed, ready for review again

PR #308, opened 2026-04-24
Fix: add support for nc files generated with chap-core v1.1.1 or earlier CLIM-599

borissim:
Updates the horizon distance values loaded from nc input files created with chap-core <= v1.1.1.
The horizon distance are increased by 1 to conform to the new 1-based indexing format.

A fixture is added that reads in a file of an older chap-core version and returns the appropriate backtest. The `plot-backtest` tests are then parametrized.


knutdrand:
A few issues to address before merging:

**1. Stray `from turtle import back` in `tests/evaluation/test_backtest_plot.py:3`**
Looks like an auto-complete accident — `turtle` is unrelated and `back` is unused. Please remove.

**2. Unnecessary conftest import in `tests/evaluation/test_backtest_plot.py:26`**
```python
from tests.evaluation.conftest import backtest, old_backtest, simulated_backtest
```
Pytest discovers conftest fixtures automatically, so this import isn't needed. It also imports a fixture named `backtest` which shadows the parametrize parameter named `backtest` in the tests below. Drop the line entirely.

**3. Unsafe attribute access in `chap_core/assessment/evaluation.py`**
```python
if Version(ds.chap_version) <= Version("1.1.1"):
```
Every other attribute read in `from_file` uses `ds.attrs.get(..., default)`. `chap_version` wasn't always written to the nc file, so files predating that attribute will raise `AttributeError` here — which is the exact class of old file this PR is trying to support. Suggest:
```python
chap_version = ds.attrs.get("chap_version", "0.0.0")
if Version(chap_version) <= Version("1.1.1"):
    forecasts_df["horizon_distance"] = forecasts_df["horizon_distance"] + 1
```
Treating a missing version as "ancient → needs the fix" matches the intent.

knutdrand:
Suggestion: apply the fix at the xarray level right after loading, rather than on the DataFrame mid-pipeline. `horizon_distance` is a dimension coord of the `forecast` DataArray, so it can be shifted directly on the Dataset:

```python
ds = xr.open_dataset(filepath)

chap_version = ds.attrs.get("chap_version", "0.0.0")
if Version(chap_version) <= Version("1.1.1"):
    ds = ds.assign_coords(horizon_distance=ds.horizon_distance + 1)

flat_data = _xarray_to_flat_data(ds)
```

Benefits:
- Single normalization point — any code path reading `ds` sees corrected values, not just the one going through `forecasts_df`.
- Backward-compat logic lives next to the load, so the flow reads as "load → normalize legacy → process" instead of a version check buried after a groupby setup.
- Easy to extend: when the next breaking change lands, add another branch here (or extract a `_normalize_legacy(ds)` helper) instead of sprinkling version checks through downstream DataFrame code.

Caveat: `assign_coords` returns a new Dataset, so rebind `ds = ds.assign_coords(...)`.

borissim:
@knutdrand (and Claude) all good stuff. I blame VS Code and copilot for the extra imports.
Ready for another around

knutdrand (approved):
Perfect

PR #316, opened 2026-04-28
refactor: updates metric plots to fit implementation strategy of back…

borissim:
New subpackage created chap_core/assessment/metric_plots/

8 individual plot files, one class each (mirroring the backtest_plots layout)

3 plots are auto-registered: metric_by_horizon_mean, metric_map, regional_metric_distribution

Updated files:

visualization.py: removed hardcoded metric_plots_registry, now uses get_metric_plots_registry() and list_metric_plots()

New docs: creating_custom_metric_plots.md — mirrors the backtest plots guide with data schema, minimal example, full example, registration instructions, and REST exposure docs

New tests: test_metric_plot.py — registry checks, decorator validation, parametrized plot rendering tests; also fixed a pre-existing test-ordering fragility in test_evaluation_plot.py where vegafusion enabled by backtest_plot.py at import time could break to_dict() calls.


borissim:
@knutdrand @mortenoh 
I Clauded this one (my first Claude PR), fixed what made sense, but I don't have the knowledge to assess the generated tests properly (whether they cover the functionality well enough). Please take a look at them with more attention.

knutdrand:
Two cleanup notes from review before merge:

**1. Import-time renderer mutation.** `chap_core/assessment/metric_plots/__init__.py:44` calls `alt.renderers.enable("browser")` at module import. That globally switches the Altair renderer for any process that imports this package — including the REST API, the CLI, and tests. The companion `backtest_plots/__init__.py` deliberately does not do this, and this PR otherwise just fixed an analogous import-time leak (vegafusion via `backtest_plot.py`) with the new autouse fixture in `test_evaluation_plot.py`. Recommend removing the `enable("browser")` call; if a script needs browser rendering, set it there.

**2. (Optional) Test registry pollution.** `test_metric_plot_decorator_sets_attributes` registers `test_decorator_plot` into the global `_metric_plots_registry` and never cleans up. `test_list_metric_plots` only asserts `>= 3`, so it tolerates the leak today, but anything that asserts on registry contents later will see it. A teardown that pops the test id (or a `monkeypatch` over the registry dict) would make the test self-contained.

borissim:
@knutdrand @mortenoh ready for (re)review

knutdrand (changes requested):
Some stuff here reverts a bit of what @mortenoh  just merged with cli perf. And there seems to be a bug in the teardown of test_metric_plot_decorator_sets_attributes.. In Claudes words: 

  Issues

  Blocker — Regresses the CLIM-631 startup-perf fix in generate_modelcard.py

  Master (post fd5d2d78, PR #318) deliberately keeps generate_modelcard.py imports lazy (function-local + TYPE_CHECKING) to keep chap CLI startup fast. This PR pulls many of those imports back up to module scope:

  chap_core/cli_endpoints/generate_modelcard.py:13-25:
  from chap_core.assessment.backtest_plots import create_plot_from_evaluation
  from chap_core.assessment.evaluation import Evaluation
  from chap_core.assessment.metrics import (..., compute_all_aggregated_metrics_from_backtest)
  from chap_core.database.model_templates_and_config_tables import ModelTemplateMetaData
  from chap_core.database.tables import BackTest
  from chap_core.external.model_configuration import ModelTemplateConfigV2
  from chap_core.plotting.evaluation_plot import make_plot_from_evaluation_object

  These are exactly the imports CLIM-631 pushed down. The if TYPE_CHECKING: block at line 27-30 still re-imports Evaluation, BackTest, ModelTemplateConfigV2 — now duplicated and dead. Also ModelTemplateMetaData was inside _placeholder_metadata_values (the comment "Defaults from ModelTemplateMetaData fields…" still
  hangs there with no preceding import).

  This file does not need to be touched by this PR at all — only the from chap_core.plotting.evaluation_plot import MetricByTimePeriodV2Mean, MetricMapV2, RegionalMetricDistributionPlot, make_plot_from_evaluation_object block inside _save_evaluation_plots needed to be split. Restore the lazy structure and only update
  those specific symbol-relocations.

  Bug — Test teardown is a no-op (and leaks _TestPlot into the registry)

  tests/evaluation/test_metric_plot.py:52:
  # tear down
  del get_metric_plots_registry()["test_decorator_plot"]

  get_metric_plots_registry() explicitly returns a copy (__init__.py:132: return _metric_plots_registry.copy()), so this del mutates a throwaway dict and the test plot stays registered for the rest of the session. It happens not to break anything today because test_registered_plot_produces_chart is parametrized at
  collection time and _TestPlot would render an empty chart anyway, but it's a latent footgun the moment any other test iterates the registry.

  Fix: expose a private _unregister(plot_id) helper or reach into the module-level _metric_plots_registry from the test (note: project convention forbids accessing private attrs from tests — so the helper is the cleaner option). Alternatively, snapshot/restore the registry with a fixture.

borissim:
@knutdrand ready for another round

knutdrand (approved):
Seems good now

PR #325, opened 2026-05-04
Stacking new

behdadnikkhah25:
### Description

This PR adds a minimal but robust stacking ensemble to CHAP, with:

- deterministic and probabilistic meta models  
- optional residual bootstrap for deterministic ensembles  
- per base model configuration via `ModelTemplate` + `ModelConfiguration`  
- a `chap evaluate-ensemble` CLI command for running ensembles on CHAP datasets  

---

#### CLI: `chap evaluate-ensemble`

- `--base-model-names`: comma separated list of base models (local or GitHub)
- `--model-configuration-yaml`: matching comma separated list of YAML configs
- `--ensemble-method`: `deterministic` or `probabilistic`
- `--random-state`: seed for ensemble RNG
- `--use-residual-bootstrap`: optional residual bootstrap (deterministic only)
- Standard dataset/backtest options
- Uses `create_model_lists` + `_TemplateWithConfig` to bind each template to its config
- Runs `Evaluation.create`, writes NetCDF + CSV reports, and global metrics

#### Training (`EnsembleModel.train`)

- Uses an inner train/validation split
- Trains all base models on inner train and predicts on validation
- Builds meta features:
  - probabilistic: aligned/resampled samples via `_SampleExtractor.reshape_samples`
  - deterministic: point forecasts via `_SampleExtractor.samples_to_flat`
- Optionally stores base residuals for bootstrap
- Fits:
  - `NonNegativeMetaModel` (NNLS) for deterministic ensembles, or
  - `ProbabilisticMetaModel` (CRPS-minimizing SLSQP) for probabilistic ensembles, with fallback to uniform weights
- Normalizes meta weights to percentages and logs them
- Retrains base models on full training data and returns an `EnsemblePredictor`

#### Prediction (`EnsemblePredictor.predict`)

- **Probabilistic:**
  - gets base samples, aligns and resamples them
  - combines them via `ProbabilisticMetaModel` into ensemble samples
  - returns a `DataSet[Samples]` aligned by `location` / `time_period`
- **Deterministic:**
  - combines base point forecasts via `NonNegativeMetaModel`
  - with residual bootstrap enabled, generates sample-level uncertainty using stored residuals and meta weights
  - otherwise returns a single deterministic forecast

---

### File overview (short)

- `chap_core/ensemble/ensemble_model.py`  
  Core stacking implementation (`EnsembleModel`) with deterministic/probabilistic modes, residual bootstrap, and RNG support. Includes legacy `EnsembleEstimator` wrapper and re-exports meta models + CRPS helpers.

- `chap_core/ensemble/_meta_models.py`  
  Meta models and CRPS utilities:
  - `NonNegativeMetaModel`: NNLS-based deterministic meta model
  - `ProbabilisticMetaModel`: CRPS-optimizing meta model over samples with robust fallback
  - `_crps_score` / `crps_ensemble`: CRPS for ensemble forecasts

- `chap_core/ensemble/_sample_extractor.py`  
  Helpers to flatten and reshape sample-based forecasts (`samples_to_flat`, `reshape_samples`) with alignment on `location` / `time_period` and controlled resampling.

- `chap_core/ensemble/_predictor.py`  
  `EnsemblePredictor` combining base predictors and the meta model at prediction time, in both deterministic and probabilistic modes.

- `chap_core/ensemble/_legacy_wrappers.py`  
  `BaseModelSpec` and `_TemplateWithConfig` to pair `ModelTemplate` with an optional config and forward it to `get_model`.

- `chap_core/cli_endpoints/ensemble.py`  
  CLI endpoint for `chap evaluate-ensemble`: loads dataset, builds template+config pairs, instantiates `EnsembleModel`, runs `Evaluation`, and writes NetCDF/CSV outputs.

---

### Tests

- `tests/ensemble/test_ensemble_model.py`
  - Probabilistic ensembles correctly reject `use_residual_bootstrap=True`.

- `tests/ensemble/test_ensemble_stacking.py`
  - Deterministic ensemble: weights sum to 100 %, output shapes are correct.
  - Deterministic + residual bootstrap: generates the requested number of samples.
  - Probabilistic ensemble: resamples to `n_samples` with correct shapes.

- `tests/ensemble/test_legacy_wrappers.py`
  - `_TemplateWithConfig` forwards config to `get_model`.
  - `BaseModelSpec` stores template/config.

- `tests/ensemble/test_meta_models.py`
  - `NonNegativeMetaModel`: non-negative coefficients, correct prediction shape.
  - `ProbabilisticMetaModel`: weights on the simplex, rejects shape mismatches, falls back to uniform weights on optimization failure.

- `tests/ensemble/test_predictor.py`
  - Deterministic `EnsemblePredictor` with bootstrap: correct sample shapes per location.
  - Probabilistic `EnsemblePredictor`: correct resampling and shapes.

- `tests/ensemble/test_sample_extractor.py`
  - `reshape_samples` respects the provided RNG.
  - Fallback behavior for `ProbabilisticMetaModel` when optimization is forced to fail.

#### Base models

Current base models:

- `rwanda_sarimax` (deterministic)  
- `chap_auto_ewars` (probabilistic)  
- `INLA_baseline_model` (probabilistic)

You can also plug in other CHAP models (e.g. `pmch`) by adding them to `--base-model-names` with a matching YAML in `--model-configuration-yaml`.

#### Dataset

- `ensemble/data/chap_LAO_admin1_monthly.csv`

---

### 1. Deterministic ensemble

```bash
chap evaluate-ensemble \
  --base-model-names \
    https://github.com/chap-models/rwanda_sarimax,\
https://github.com/chap-models/chap_auto_ewars,\
https://github.com/chap-models/INLA_baseline_model \
  --model-configuration-yaml \
    ensemble/config/rwanda_sarimax_config.yaml,\
ensemble/config/chap_ewars_monthly_config.yaml,\
ensemble/config/inla_baseline_config.yaml \
  --ensemble-method deterministic \
  --use-residual-bootstrap \
  --random-state 20 \
  --dataset-csv ensemble/data/chap_LAO_admin1_monthly.csv \
  --report-filename ensemble_report.csv \
  --output-file ensemble_report.nc \
  --backtest-params.n-splits 3 \
  --backtest-params.n-periods 3

```
---

### 2. Probabilistic ensemble

```bash
chap evaluate-ensemble \
  --base-model-names \
    https://github.com/chap-models/rwanda_sarimax,\
https://github.com/chap-models/chap_auto_ewars,\
https://github.com/chap-models/INLA_baseline_model \
  --model-configuration-yaml \
    ensemble/config/rwanda_sarimax_config.yaml,\
ensemble/config/chap_ewars_monthly_config.yaml,\
ensemble/config/inla_baseline_config.yaml \
  --ensemble-method probabilistic \
  --random-state 20 \
  --dataset-csv ensemble/data/chap_LAO_admin1_monthly.csv \
  --report-filename ensemble_report.csv \
  --output-file ensemble_report.nc \
  --backtest-params.n-splits 3 \
  --backtest-params.n-periods 3

```

---

### Notes

The following files are kept only as historical/experimental references from earlier iterations and are not part of the new ensemble API:

- chap_core/ensemble/stackedEnsemble.py
- chap_core/ensemble/ensemble_model_v2.py
- chap_core/ensemble/debug_prob_meta.py
- chap_core/cli_endpoints/ensemble_v2.py


knutdrand:
_This comment was generated by Claude. It covers **housekeeping / hygiene only** — a separate review of the ensemble business logic (CRPS minimization, stacking split, residual bootstrap, weight normalization) will follow. Note that the `# type: ignore` annotations and the underscore-prefixed names re-exported in `__all__` make it harder to spot real logic issues, so cleaning those up first will make the business-logic review more reliable._

## Blocking

1. **Dead/experimental code committed.** The PR description marks these as historical references:
   - `chap_core/ensemble/classes/stackedEnsemble.py`
   - `chap_core/ensemble/ensemble_model_v2.py`
   - `chap_core/ensemble/debug_prob_meta.py`
   - `chap_core/ensemble/run_debug_prob_meta.py`
   - `chap_core/cli_endpoints/ensemble_v2.py`

   They are silenced in `pyproject.toml` (mypy ``ignore_errors = true``, pyright ``exclude``, ruff ``[\"ALL\"]``) rather than fixed. Delete them; revisit in a separate PR if needed.

2. **Runtime artifacts and bulk test data committed.**
   - `chap_core/ensemble/data/chap_LAO_admin1_monthly.csv` (2809 lines) — only referenced from README examples.
   - `chap_core/mlflow.db` — binary SQLite, almost certainly accidental.
   - `chap_core/ensemble_meta_report.csv`, `chap_core/metrics_comparison.csv`, `chap_core/metrics_comparison_sjekk.csv` — debug/scratch output.

3. **`EnsembleModel.train` writes `ensemble_meta_report.csv` to CWD as a side effect** (`ensemble_model.py:4358-4362`), with the write duplicated again in `EnsembleEstimator.train`. This is also how the file in (2) got committed. Training should not write to arbitrary paths; move any report emission to the CLI layer.

4. **CLI help strings are in Norwegian** (`chap_core/cli_endpoints/ensemble.py:287-326`). The rest of the CLI is English; ``chap evaluate-ensemble --help`` should match. Same for inline comments in `_legacy_wrappers.py`, `_sample_extractor.py`, and `ensemble.py`.

5. **No test for the new CLI / `_evaluate_ensemble_core`.** Existing tests cover the model classes with dummy estimators only. CLAUDE.md rule 9 requires a test for new features — at minimum a smoke test running the command against a small fixture.

6. **`chap_core/ensemble/__init__.py` is just `# test`.** Replace with empty file or proper docstring.

## Significant

7. **Underscore names in public `__all__`.** `ensemble.py` exports `_evaluate_ensemble_core` and `_save_reports`; `ensemble_model.py` re-exports `EnsemblePredictor` from a `_predictor.py` private module. Pick one convention. (This is the kind of thing that masks real issues — code reviewers can't tell what's API vs internal.)

8. **Unrelated changes mixed in.** `pytz` typing fix in `time_period/date_util_wrapper.py`, and `_common.py:save_results` switched from headerless CSV to `header=True` — a behavior change to the existing `chap evaluate` output. Split out.

9. **`_compute_metrics` swallows all `Exception`s per metric** (`ensemble.py:148-155`). Hides shape/type bugs under \"metric failed\". Narrow the catch.

10. **`_TemplateWithConfig.__getattr__`** forwards everything to `self._template` (`_legacy_wrappers.py:559-560`). Risks recursion if `_template` is unset and defeats type checking — prefer explicit forwarding for the attributes actually used (`name`, `repo`).

11. **`# type: ignore` escape hatches** scattered through `ensemble_model.py` and `_predictor.py` (`[call-arg]`, `[arg-type]`, `[misc]`). These are likely hiding real protocol/typing mismatches in the predictor + meta-model interface that would otherwise show up as bugs.

## Smaller things

- `_meta_models.py`: `n = len(X_samples)` set twice (lines 654, 672).
- `_predictor.py` residual bootstrap loops in pure Python over `(model_idx, row_idx)` — vectorize with `rng.choice(res_clean, size=(n_rows, s_count))`.
- `_predictor.py:780-792` (deterministic non-bootstrap): sorts by key cols, but assumes `tp = future_data[loc].time_period` matches `df_loc[\"forecast\"]` ordering — only a length check guards this.
- `ensemble.py:_save_reports` accepts a `forecasts` arg it immediately `del`s.
- `ensemble.py:296`: `polygons_id_field: str | None = \"id\"` is awkwardly typed nullable with non-None default.
- `EnsembleEstimator.train` duplicates the parent's meta-report write.

## Test coverage gaps
- No test exercises the NaN masking branch in `EnsembleModel.train`.
- No test for the `s <= 0` fallback in `_predictor.py`.
- Dummy estimator/template classes duplicated between `test_ensemble_stacking.py` and `test_predictor.py` — extract to a shared conftest (CLAUDE.md rule 12).

## Recommendation
Address the blocking items (delete dead code/artifacts, remove the train-time CSV write, translate CLI help, add a CLI test, fix `__init__.py`) and clear out the `# type: ignore` annotations + underscore re-exports before the business-logic review.

behdadnikkhah25:
## Ensemble fix: finalize stacking pipeline, cleanup, and CLI reporting

### Summary

This PR finalizes the stacking ensemble in CHAP and addresses the previous review comments:

- deterministic and probabilistic meta-models  
- optional residual bootstrap for deterministic ensembles  
- per–base-model config via `ModelTemplate` + `ModelConfiguration`  
- `chap evaluate-ensemble` CLI for running ensembles on CHAP datasets (incl. URL datasets)  
- cleaned-up internals (no dead/legacy code, no `type: ignore`), clearer exports, and tests  

---

### High-level design

#### CLI: `chap evaluate-ensemble`

- `--base-model-names`: comma-separated base models (local or GitHub URLs)  
- `--model-configuration-yaml`: matching comma-separated YAML configs  
- `--ensemble-method`: `"deterministic"` or `"probabilistic"`  
- `--use-residual-bootstrap`: residual bootstrap for deterministic only (rejected for probabilistic)  
- `--random-state`: seed for ensemble RNG  

**Dataset options:**

- `--dataset-name` / `--dataset-country` (built-in datasets)  
- `--dataset-csv` (local path or HTTP(S) URL, without forcing `Path` on URLs)  

**Evaluation:**

- uses `create_model_lists` + `_TemplateWithConfig` to bind template+config  
- runs `Evaluation.create(...)`, writes:
  - NetCDF (`.nc`)
  - metrics CSV (`ensemble_report.csv`) with metric names + `ensemble_model`, `model_name`, `ensemble_method`
  - base-model weights (`ensemble_meta_report.csv`)  

---

### Training & prediction overview

#### `EnsembleModel.train`

- inner train/validation split on `time_period`  
- trains base models on inner train, predicts on validation  

**Builds meta-features:**

- probabilistic: `_SampleExtractor.reshape_samples` → `[n_val, n_samples]` per model  
- deterministic: `_SampleExtractor.samples_to_flat` → `[n_val, n_models]`  

**Optional residual bootstrap:**

- collects/cleans residuals per base model for later sampling  

**Meta-models:**

- deterministic: `NonNegativeMetaModel` (NNLS, non-negative simplex weights)  
- probabilistic: `ProbabilisticMetaModel` (CRPS minimization on simplex, with uniform fallback)  

**Weights:**

- normalizes `coef_` to percent (`self.weights`) and logs them  

**Finalization:**

- retrains base models on full data  
- returns an `EnsemblePredictor` with:
  - base predictors  
  - fitted meta-model  
  - method  
  - `n_samples`  
  - bootstrap config  
  - RNG  

---

#### `EnsemblePredictor.predict`

**Probabilistic:**

- aligns/resamples base samples via `_SampleExtractor.reshape_samples`  
- combines with `ProbabilisticMetaModel.predict`  
- returns `DataSet[Samples]` per location/time period  

**Deterministic:**

- stacks point forecasts, uses `NonNegativeMetaModel.predict`  

With bootstrap:

- samples residuals per model  
- builds non-negative samples  
- combines with meta-weights  
- returns `DataSet[Samples]` with `n_samples`  

Without bootstrap:

- returns `DataSet[Samples]` with a single `sample_0` per time step  
- explicit checks for row/time-period mismatches per location  

---

### File overview (short)

#### `chap_core/ensemble/ensemble_model.py`

- `EnsembleModel`: deterministic/probabilistic stacking, optional residual bootstrap, RNG, learned weights  
- `EnsembleEstimator`: legacy wrapper around `EnsembleModel`  
- re-exports: `BaseModelSpec`, `EnsemblePredictor`, `NonNegativeMetaModel`, `ProbabilisticMetaModel`, `crps_ensemble`  

---

#### `chap_core/ensemble/_meta_models.py`

- `_crps_score`, `crps_ensemble`  
- `NonNegativeMetaModel`: NNLS, non-negative normalized weights  
- `ProbabilisticMetaModel`: CRPS minimization on simplex, shape checks + uniform fallback  

---

#### `chap_core/ensemble/_sample_extractor.py`

- `samples_to_flat`: `Samples` → `location`, `time_period`, `forecast`  
- `reshape_samples`: aligns on `location,time_period`, resamples to target `n_samples` using RNG  

---

#### `chap_core/ensemble/_predictor.py`

- `EnsemblePredictor`: prediction-time combination (deterministic/probabilistic + optional bootstrap)  

---

#### `chap_core/ensemble/_legacy_wrappers.py`

- `BaseModelSpec`: template + optional config  
- `_TemplateWithConfig`: forwards config to `get_model`, exposes `name` / `repo`  

---

#### `chap_core/cli_endpoints/ensemble.py`

`evaluate-ensemble` CLI:

- loads data (`--dataset-name` or `--dataset-csv` including URLs)  
- builds template+config pairs  
- runs `Evaluation`  
- writes:
  - `.nc`
  - `ensemble_report.csv` (metrics with headers + ensemble metadata)
  - `ensemble_meta_report.csv` (base-model weights)  

---

### Tests (short)

#### `tests/ensemble/test_ensemble_model.py`

- probabilistic ensembles reject `use_residual_bootstrap=True`  
- weights normalized and stored as percents  

---

#### `tests/ensemble/test_ensemble_stacking.py`

- deterministic: weights sum to 100 %, shapes correct  
- deterministic + bootstrap: correct sample counts  
- probabilistic: resampling and shapes correct  

---

#### `tests/ensemble/test_legacy_wrappers.py`

- `_TemplateWithConfig` forwards config  
- `BaseModelSpec` stores template/config  

---

#### `tests/ensemble/test_meta_models.py`

- `NonNegativeMetaModel`: non-negative, correct shapes  
- `ProbabilisticMetaModel`: simplex weights, shape validation, uniform fallback  

---

#### `tests/ensemble/test_predictor.py`

- deterministic + bootstrap: correct shapes, non-negative samples  
- probabilistic: resampling behavior and shapes  
- clear failure on misaligned base predictions  

---

#### `tests/ensemble/test_sample_extractor.py`

- `reshape_samples` respects RNG and alignment  
- `samples_to_flat` fallbacks (`forecast` / `value` / `sample_*`)  

---

#### `tests/cli_endpoints/test_ensemble_cli.py`

- end-to-end smoke test for `evaluate-ensemble` / `_evaluate_ensemble_core`  
- verifies `use_residual_bootstrap` propagation and report generation  

---

#### `Tests are run with`
- `uv run make lint` (for detecting bugs in the files)
- `uv run make test` (running all tests for ensemble)

--- 

### Example usage (fixed code)

#### Base models:

- rwanda_sarimax (deterministic)
- chap_auto_ewars (probabilistic)
- INLA_baseline_model (probabilistic)

#### Dataset:

- No local CSV in the repo; use URL via --dataset-csv.

---

### 1) Deterministic ensemble (with residual bootstrap)

```bash
chap evaluate-ensemble \
  --base-model-names https://github.com/chap-models/rwanda_sarimax,https://github.com/chap-models/chap_auto_ewars,https://github.com/chap-models/INLA_baseline_model \
  --model-configuration-yaml ensemble/config/rwanda_sarimax_config.yaml,ensemble/config/chap_ewars_monthly_config.yaml,ensemble/config/inla_baseline_config.yaml \
  --ensemble-method deterministic \
  --use-residual-bootstrap \
  --random-state 20 \
  --dataset-csv https://raw.githubusercontent.com/dhis2/climate-health-data/main/lao/chap_LAO_admin1_monthly.csv \
  --report-filename ensemble_report.csv \
  --output-file ensemble_report.nc \
  --backtest-params.n-splits 3 \
  --backtest-params.n-periods 3

```

(For a purely point-based deterministic ensemble, remove --use-residual-bootstrap.)

---

### 2) Probabilistic ensemble

```bash
chap evaluate-ensemble \
  --base-model-names https://github.com/chap-models/rwanda_sarimax,https://github.com/chap-models/chap_auto_ewars,https://github.com/chap-models/INLA_baseline_model \
  --model-configuration-yaml ensemble/config/rwanda_sarimax_config.yaml,ensemble/config/chap_ewars_monthly_config.yaml,ensemble/config/inla_baseline_config.yaml \
  --ensemble-method probabilistic \
  --random-state 20 \
  --dataset-csv https://raw.githubusercontent.com/dhis2/climate-health-data/main/lao/chap_LAO_admin1_monthly.csv \
  --report-filename ensemble_report.csv \
  --output-file ensemble_report.nc \
  --backtest-params.n-splits 3 \
  --backtest-params.n-periods 3

```

knutdrand:
## Two places where chap-core already has what you need

Looking through the implementation I noticed two spots that re-implement existing chap-core functionality. Each is defensible in isolation, but together they're worth raising — partly because both are easy wins, and partly because the underlying habit (auditing what already exists before building) is the most useful engineering lesson I can think of for working in a shared codebase.

### CRPS

You implement `_crps_score` and `crps_ensemble` in `_meta_models.py`. chap-core already has `CRPSMetric` in `chap_core/assessment/metrics/crps.py`, and the evaluation pipeline already reports it on the ensemble's output.

Worth knowing: your version is actually the better implementation. `O(m log m)` via sort+cumsum instead of the naive `O(m²)` pairwise matrix in `CRPSMetric.compute_sample_metric`, and the unbiased "fair" estimator (`1/(m(m-1))`) instead of the biased one (`1/m²`). The unbiased form is the textbook recommendation.

The cost of having both: the meta-model is trained by minimizing one CRPS estimator, and the CLI reports a different one at evaluation time. They agree in expectation but differ by `O(1/m)`.

What I'd suggest: lift `_crps_score` into `chap_core/assessment/metrics/crps.py` (replacing the naive body of `compute_sample_metric`) and import from there in the meta-model. One CRPS in the codebase, your better implementation lands where everyone benefits. If you'd rather not touch `CRPSMetric`, at minimum add a docstring on `_crps_score` saying "unbiased `O(m log m)` form; `CRPSMetric` uses the biased `O(m²)` form; equivalent in expectation" — so the next reader doesn't have to reconstruct the analysis.

### Time split for meta-features

In `EnsembleModel.train` you build the inner train/val split by stringifying periods, sorting alphabetically, and masking the dataframe:

```python
all_periods = sorted(df["time_period"].dropna().astype(str).unique())
split_idx = ... len(all_periods) - self.inner_val_periods
```

chap-core has `train_test_split` and `train_test_generator` in `chap_core/assessment/dataset_splitting.py`. They operate on `TimePeriod` objects (proper ordering, no reliance on string sort) and use `restrict_time_period` to preserve dataset structure. They're used throughout the codebase — including by the outer evaluation that calls into your ensemble — so re-implementing them inside the ensemble creates an inconsistency.

Two suggestions:

- **For the single split**, use `train_test_split(train_data, split_period)`. That removes the `to_pandas → mask → from_pandas(..., fill_missing=True)` round-trip. The `fill_missing=True` is quietly inserting rows for absent (location, period) pairs — probably not what you want for a meta-feature matrix.
- **More ambitiously**, switch to `train_test_generator` so the meta-model fits on multiple expanding-window splits instead of one. With `inner_val_periods=12` and a handful of locations you have at most ~60 examples to fit `k` weights right now. Blocked CV is a small loop change and would give you 5–10× the meta-features. The thing you got right here — putting val data temporally *after* train data — is the hard part. Going from one block to N blocks of the same shape is mechanical.

The string-sort is also a latent bug: fine for ISO period IDs (`"2024-01" < "2024-02"`) but would silently misorder anything else. Using `TimePeriod` objects avoids it entirely.

### The habit behind this

Before adding new infrastructure to a shared codebase, the question worth spending an hour on is *"does this already exist, and if so, why isn't it good enough?"* — and writing the answer down. That's true whether the existing thing is something you should reuse as-is or something you have a good reason to improve on. It protects against rediscovering subtle bugs others already fixed, makes design choices reviewable on their merits, and gives the next reader a starting point.

Neither item is blocking — but both should be either replaced with the existing helpers or have a short comment explaining why they aren't.

knutdrand:
## The probabilistic combination scheme needs a written design decision (and probably a fix)

This is a separate concern from the chap-core reuse points: there's a design choice inside `ProbabilisticMetaModel` that needs a written rationale, and once you look at what it actually does, also needs to change.

### What the code does

`ProbabilisticMetaModel.predict` (and the objective in `fit`) computes

```python
ens = sum(w[i] * X_samples[i] for i in range(len(X_samples)))
```

column-wise. So "ensemble sample `k`" = `w_0 · base_0_sample_k + w_1 · base_1_sample_k + …`.

The problem: base models emit Monte Carlo samples in whatever order their internal RNG produced. Column `k` of base A and column `k` of base B aren't coupled samples — they're arbitrarily aligned. So the resulting "distribution" depends on incidental sample ordering across base models.

The variance is also artificially low. A sum of weighted (effectively independent) samples has variance `Σ w_i² σ_i²`, much smaller than a proper mixture's `Σ w_i σ_i² + Σ w_i (μ_i − μ̄)²`. The CRPS optimization then picks weights that look good for *this* narrow distribution — which is not the distribution you'd want at inference time.

### The two standard alternatives

Both are well-known in the forecast-combination literature, and both are 1–3 line changes:

- **Linear pool / mixture.** For each ensemble sample `k`, draw a base index `i_k ~ Categorical(w)` and emit that base's sample. Produces a proper mixture distribution — the default in the literature.

  ```python
  idx = rng.choice(n_bases, size=n_samples, p=self.coef_)
  ens_samp[:, k] = X_samples[idx[k]][:, k]
  ```

- **Vincentization (quantile averaging).** Sort each base's samples per row first, then column-wise weighted sum. Sharper than the mixture, equivalent to averaging quantile functions.

  ```python
  mat_sorted = np.sort(mat, axis=1)
  ```

Either gives the CRPS optimization something coherent to optimize over.

### What I'd want to see

Pick one of the two schemes and add a short docstring on `ProbabilisticMetaModel` explaining which you picked and why. Right now the PR description, the class docstring, and the inline comments are all silent on the combination scheme — a reader has to read the array math and infer, and the inference is "this isn't what was meant."

The good parts of this code — CRPS optimization over the simplex with SLSQP, the uniform-weight fallback, the simplex constraints, the vectorized CRPS itself — are solid. They deserve a clear statement of what they're optimizing for.

### The thing to take away

Unlike the CRPS / time-split comment, this isn't "the helper already exists, reuse it" — there's no canonical chap-core helper here because the design question is genuinely open. Several reasonable people could pick different combination schemes for defensible reasons. **The part that's not optional is writing down which one you picked.** A two-sentence docstring saying "we use a linear pool because …" turns this from something a reviewer has to guess at into a deliberate design choice that can be challenged on its own terms. Naming and justifying the design choices you make — especially the ones where the literature doesn't have a single right answer — is one of the most useful habits to develop heading into research-adjacent work.

behdadnikkhah25:
## Align ensemble CRPS, time splitting, and probabilistic combining

### Summary

This PR addresses the two main review comments:

- reuse existing chap-core helpers (CRPS and time-based splitting), and  
- make the probabilistic combination scheme both correct and explicitly documented (vincentization).

It also expands test coverage to validate the new behaviour end-to-end.

---

### Changes since last review

- Consolidated CRPS into `chap_core/assessment/metrics/crps.py` and reuse it from the ensemble meta-model.
- Switched the internal time split in `EnsembleModel.train` to `train_test_split` based on `TimePeriod`.
- Replaced the probabilistic combination with vincentization and documented the design choice.
- Added tests for:
  - vincentization invariance to sample ordering
  - unbiased CRPS matrix helper
  - time-split logic
  - monotone/sorted probabilistic ensemble samples.

---

### Design changes related to review feedback

#### 1) Reuse of chap-core functionality (CRPS + time split)

**CRPS**

- The CRPS implementation has been consolidated in `chap_core/assessment/metrics/crps.py` as:
  - `crps_score_unbiased(samples: np.ndarray, observed: float) -> float` and
  - `crps_score_unbiased_matrix(observations: np.ndarray, forecasts: np.ndarray) -> float`.
- This uses the unbiased (“fair”) estimator with factor `1 / (m * (m - 1))` over `m` samples and an `O(m log m)` algorithm based on sorting and cumulative sums, replacing the previous naive `O(m^2)` pairwise implementation.
- `CRPSMetric` and `CRPSLog1pMetric` now delegate to `crps_score_unbiased`, and the probabilistic ensemble meta-model uses `crps_score_unbiased_matrix`. Training and evaluation now use the same CRPS definition.

**Time-based split**

- `EnsembleModel.train` no longer reimplements its own time split via string-sorting and pandas masking.
- It now uses `train_test_split` from `chap_core.assessment.dataset_splitting` directly, operating on `TimePeriod` objects.
- This removes reliance on alphabetical string ordering and avoids silent `fill_missing=True` behaviour when reconstructing the dataset.

#### 2) Probabilistic combination with an explicit design choice

- The probabilistic meta-model now combines base predictive distributions via **vincentization (quantile averaging)**:
  - for each location/horizon, it sorts each base model’s samples per row to approximate their quantile functions $Q_i(p)$,
  - then forms ensemble quantiles as
    ```math
    Q_{\text{ens}}(p) = \sum_i w_i Q_i(p),
    ```
    where $w_i$ are non-negative weights on the simplex.
- This makes the ensemble distribution:
  - **invariant to the internal sample ordering** of each base model, and
  - typically sharper than a pure mixture, while remaining a standard construction in forecast combination.
- The choice of vincentization (rather than linear pooling of raw samples) is now documented in the `ProbabilisticMetaModel` docstring.

**Behavioural note**

- The probabilistic ensemble now represents the forecast distribution via **vincentized quantiles**:
  - `samples` per row are sorted quantiles, not unordered Monte Carlo draws.
  - This yields **monotonically non-decreasing** (often partially flat) sample sequences per row, which is expected and desired in a quantile-based representation.
- Compared to the previous implementation (which linearly combined raw samples column-wise and tended to produce artificially narrow distributions), CRPS may change:
  - the new scheme optimizes an unbiased CRPS estimator for a more realistic uncertainty distribution,
  - so scores are more “honest” about spread, even if they are not directly comparable to earlier biased / overly sharp ensembles.
- Deterministic ensembles keep their previous behaviour; they are only affected indirectly in that evaluation now uses the unified unbiased CRPS implementation for metrics.

---

### File overview

#### `chap_core/assessment/metrics/crps.py`

- Adds:
  - `crps_score_unbiased(samples: np.ndarray, observed: float) -> float`
  - `crps_score_unbiased_matrix(observations: np.ndarray, forecasts: np.ndarray) -> float`
- Both use the unbiased estimator with `1 / (m * (m - 1))` and an `O(m log m)` sort + cumsum algorithm.
- `CRPSMetric` and `CRPSLog1pMetric` now delegate to `crps_score_unbiased`.

#### `chap_core/ensemble/_meta_models.py`

- `crps_ensemble` now delegates to `crps_score_unbiased_matrix`.
- Introduces `_vincentize_samples(X_samples, weights)`:
  - validates shapes,
  - stacks base samples as `(n_models, n_rows, n_samples)`,
  - sorts along the sample axis and combines with simplex weights to produce vincentized ensemble samples.
- `ProbabilisticMetaModel`:
  - uses `_vincentize_samples` in both `fit` and `predict`,
  - optimizes CRPS over the simplex via SLSQP,
  - has a clear docstring explaining vincentization and invariance to sample ordering,
  - includes a robust fallback to uniform weights if optimization fails or returns invalid coefficients.
- `NonNegativeMetaModel` unchanged in behaviour, still providing non-negative weights with NNLS.

#### `chap_core/ensemble/ensemble_model.py`

- `EnsembleModel.train`:
  - uses `train_test_split` from `chap_core.assessment.dataset_splitting` to create inner train/validation splits based on `TimePeriod`,
  - removes custom string-based sorting and pandas masking for splits.

_(Other ensemble internals are unchanged except for wiring into the new helpers.)_

---

### Tests

#### `tests/ensemble/test_meta_models.py`

- **Simplex constraints**: `test_probabilistic_meta_model_weights_on_simplex`  
  Ensures `ProbabilisticMetaModel.coef_` is non-negative and sums to 1.
- **Shape validation**: `test_probabilistic_meta_model_rejects_shape_mismatch`  
  Validates that mismatched sample shapes raise a `ValueError`.
- **Vincentization invariance**: `test_probabilistic_meta_model_invariant_to_sample_order`  
  Confirms that permuting sample columns per base model does not change ensemble predictions.
- **Optimization fallback**: `test_probabilistic_meta_model_fallback_on_failed_opt`  
  Forces a failed optimization and checks fallback to uniform weights.

#### `tests/ensemble/test_ensemble_model.py`

- **Input validation**:
  - `test_probabilistic_disallows_residual_bootstrap`: probabilistic ensembles reject `use_residual_bootstrap=True`.
  - `test_requires_base_templates`: at least one base model is required.
  - `test_invalid_method_raises`: invalid `method` values raise.
- **Split requirements**:
  - `test_train_requires_two_periods`: at least two time periods are needed.
  - `test_train_invalid_split_raises`: invalid `inner_val_periods` raises a clear error.
- **TimePeriod-based split**:
  - `test_train_uses_timeperiod_split`: monkeypatches `train_test_split` and verifies:
    - the correct `prediction_start_period` is passed based on `inner_val_periods`,
    - it is a `TimePeriod`, not a string.

#### `tests/ensemble/test_ensemble_stacking.py`

- **Deterministic ensembles**:
  - `test_deterministic_predict_shape_and_weights`: weights sum to 100%, shapes as expected.
  - `test_deterministic_residual_bootstrap_generates_samples`: correct `n_samples` with bootstrap.
  - `test_deterministic_residual_bootstrap_varies_samples`: bootstrap actually introduces variation.
- **Probabilistic ensembles**:
  - `test_probabilistic_predict_samples_count`: correct sample count and shape for probabilistic ensembles.
  - `test_probabilistic_ensemble_outputs_sorted_samples`: each row’s samples are monotonically non-decreasing (vincentized quantiles).

#### `tests/ensemble/test_predictor.py`

- Deterministic path:
  - bootstrap sample counts and shapes,
  - weight fallback when meta weights are all zero,
  - failure if base predictors return misaligned rows.
- Probabilistic path:
  - correct sample counts and shapes when combining base probabilistic predictors.

#### `tests/ensemble/test_sample_extractor.py`

- `reshape_samples`:
  - respects RNG,
  - preserves alignment with reference `(location, time_period)` pairs.
- Includes a test that `ProbabilisticMetaModel` falls back to uniform weights if optimization fails (via monkeypatched `minimize`).

#### `tests/ensemble/test_legacy_wrappers.py`

- `_TemplateWithConfig` forwards configuration into `get_model`.
- `BaseModelSpec` stores and exposes `template` and `config`.

---

#### `Tests are run with`
- `uv run make lint` (for detecting bugs in the files)
- `uv run make test` (running all tests for ensemble)

---

### Example usage

#### 1) Deterministic ensemble (with residual bootstrap)

```bash
chap evaluate-ensemble \
  --base-model-names https://github.com/chap-models/rwanda_sarimax,https://github.com/chap-models/chap_auto_ewars,https://github.com/chap-models/INLA_baseline_model \
  --model-configuration-yaml ensemble/config/rwanda_sarimax_config.yaml,ensemble/config/chap_ewars_monthly_config.yaml,ensemble/config/inla_baseline_config.yaml \
  --ensemble-method deterministic \
  --use-residual-bootstrap \
  --random-state 20 \
  --dataset-csv https://raw.githubusercontent.com/dhis2/climate-health-data/main/lao/chap_LAO_admin1_monthly.csv \
  --report-filename ensemble_report.csv \
  --output-file ensemble_report.nc \
  --backtest-params.n-splits 3 \
  --backtest-params.n-periods 3

```

#### 2) Probabilistic ensemble
```bash
chap evaluate-ensemble \
  --base-model-names https://github.com/chap-models/rwanda_sarimax,https://github.com/chap-models/chap_auto_ewars,https://github.com/chap-models/INLA_baseline_model \
  --model-configuration-yaml ensemble/config/rwanda_sarimax_config.yaml,ensemble/config/chap_ewars_monthly_config.yaml,ensemble/config/inla_baseline_config.yaml \
  --ensemble-method probabilistic \
  --random-state 20 \
  --dataset-csv https://raw.githubusercontent.com/dhis2/climate-health-data/main/lao/chap_LAO_admin1_monthly.csv \
  --report-filename ensemble_report.csv \
  --output-file ensemble_report.nc \
  --backtest-params.n-splits 3 \
  --backtest-params.n-periods 3

```

chap_core/cli_endpoints/_common.py
knutdrand, line 1: Ideally should be no changes in core code. Document any changes here well, or revert

chap_core/assessment/evaluation.py
knutdrand, line 1: Ideally should be no changes in core code. Document any changes here well, or revert

tests/fixtures/ensemble_config/chap_ewars_monthly_config.yaml
knutdrand, line 1: These config files should not be in main repo, move to test fixtures if they are needed by tests

chap_core/ensemble/wrappers.py
knutdrand, line 1: Legacy indicates not longer in use, if so delete

knutdrand:
Må rebases på master før merge, så prøv å reduser kodeforandwinger i hoved_repo til et minimum. Så tror jeg cli kommandoen treneger en liten cleanup sånn at den matcher eksisterende kommandoer, men det kan gjøres i et nytt pr. Jeg er litt usikker på den deterministiske implementasjonen. Føler kanskje ikke bootstrap hører hjemme der, at det er bedre at det er en ren deterministisk modell? Er litt implementasjonsting fra Claude der og, kommer i en ny kommentar. Her er jeg usikker på hva som er riktig implementasjon, men når du får igang et bra analysescript kan du jo prøve å kjøre før og etter endringene han foreslår

knutdrand:
Og bruk median ikke mean når du går fra probabilistiske sampler til ett determinstisk tall

knutdrand:
1. NNLS coefficients are renormalized to the simplex after fitting — this changes what they predict

  chap_core/ensemble/_meta_models.py (NonNegativeMetaModel.fit):

  coef_raw, _ = nnls(X, y)
  coef = np.asarray(coef_raw, float)
  s = coef.sum()
  self.coef_ = coef / s if s > 0 else coef

  NNLS solves argmin_β ||Xβ − y||² s.t. β ≥ 0. The minimizer can have sum(β) ≠ 1 — and that scale is meaningful. If all base models are systematically biased
  low, NNLS will respond with sum(β) > 1 to scale up; if biased high, sum(β) < 1. Renormalizing destroys that adjustment, and predict() then returns X @ (β/s),
   which is not the NNLS solution.

  Worked example: if NNLS gives β = [0.8, 0.4] (because bases are biased low and 1.2 · ⟨base⟩ matches y), the renormalized prediction is 0.667 · X_0 + 0.333 ·
  X_1 — systematically biased low at predict time.

  Pick one of:
  - Keep raw NNLS coefficients for predict(). Use the simplex form only for the reported self.weights (the percentages logged for the user). Two separate
  quantities.
  - Constrain the fit to the simplex with SLSQP (sum(w) = 1, w ≥ 0) so the optimizer is solving the actual problem you want.

  Either is fine — but the current code optimizes one objective and uses a different solution at predict time. That's a silent bias.

  Related: NonNegativeMetaModel fits no intercept. With non-negative weights and no intercept, the only way the meta-model can absorb constant base-model bias
  is by pushing sum(β) away from 1 — which the renormalization then erases. Adding an intercept (or constraining the simplex inside the fit) makes this
  consistent.

  ---
  2. Residual bootstrap has the same pathology vincentization fixed on the probabilistic side

  chap_core/ensemble/_predictor.py:71-83 builds ensemble samples like this:

  for model_idx, residuals in enumerate(self._base_residuals):
      ...
      sampled_res = self._rng.choice(res_clean, size=(n_rows, s_count), replace=True)
      model_samples = np.maximum(base_pred + sampled_res, 0.0)
      ens_samp += w[model_idx] * model_samples

  So ensemble sample k at row r is Σ_i w_i · (base_i_pred + ε_{i,k}), where each ε_{i,k} is drawn independently from base i's residual distribution. The
  implicit assumption is that base errors are uncorrelated across models. They aren't — bases trained on the same data tend to err in the same direction on
  hard rows. A sum of independently sampled residuals has variance Σ w_i² σ_i², much smaller than the true Σ w_i σ_i² + 2 Σ_{i<j} w_i w_j Cov(ε_i, ε_j). The
  bootstrap distribution will be artificially narrow — exactly the pattern you flagged for the column-wise probabilistic combination.

  The standard fix is simpler than the per-base scheme: do residual bootstrap on the ensemble prediction directly.

  # at train time, after meta_model.fit():
  ens_pred_val = meta_det.predict(X_clean_mat)
  self._ensemble_residuals = y_clean - ens_pred_val      # one residual per row

  # at predict time:
  sampled_res = rng.choice(self._ensemble_residuals, size=(n_rows, s_count))
  ens_samp = np.maximum(y_point[:, None] + sampled_res, 0.0)

  This is the textbook residual bootstrap for a regression model. It captures whatever joint error structure the bases actually have (because the inner-val
  residuals are observed, not modeled), it's simpler, it removes the broken independence assumption, and the resulting bands are correctly calibrated against
  in-sample MAE.

  If you want to keep per-base residuals for a different reason, the alternative is to bootstrap whole residual rows: treat the [n_val × n_bases] matrix of
  inner-val residuals as one block, draw row indices r_k ~ Uniform(1..n_val), and emit Σ_i w_i · (base_i_pred + R[r_k, i]). That preserves cross-base
  correlation.

  ---
  3. Inner-val residuals come from under-trained base models, but are used to widen predictions from fully retrained bases

  In train():

  - residuals are collected from preds_inner = [e.train(inner_train) for e in ests] — bases trained on roughly (N − inner_val_periods)/N of data.
  - after meta-fit, full_predictors = [e.train(train_data) for e in full_ests] retrains bases on all data.
  - EnsemblePredictor.predict then adds the inner-trained residuals to predictions from the full-trained bases.

  The retrained bases are more accurate, so their residual distribution should be narrower. Using the larger inner-trained residuals systematically over-widens
   the bootstrap intervals.

  This is the textbook "stacking on a single split" pitfall. Cleanest fix is what you flagged in the earlier comment as the "more ambitious" version:
  train_test_generator for K-fold out-of-fold residuals — same model size as the inner case but full data coverage, and the gap to the retrained predictor is
  smaller and more consistent.

  Even without restructuring the split, the issue should be documented in the ProbabilisticMetaModel-style design-decision docstring. The "deterministic +
  residual bootstrap" ensemble is doing more than the name suggests — it's a stacking ensemble whose uncertainty quantification has a known calibration bias
  unless this is addressed.

  ---
  Smaller things worth flagging

  - samples_to_flat collapses probabilistic base outputs to their mean (chap_core/ensemble/_sample_extractor.py:23-25, the df[sample_cols].mean(axis=1)
  fallback). Silent — a user running --ensemble-method deterministic with probabilistic base models loses sample information without warning. Log when this
  fallback fires, per base.
  - NaN-row drop is silent (ensemble_model.py:127-135). nan_in_features = np.any(np.isnan(meta_mat), axis=1) can quietly halve the effective fit size. Log
  dropped count + per-base NaN counts.
  - No re-centering of residuals. Standard residual bootstrap subtracts the mean before resampling. Combined with the issues in (1) and (3), the residual mean
  isn't zero, so the bootstrap shifts the ensemble forecast by mean(residuals).
  - np.maximum(base_pred + sampled_res, 0.0) per base, before weighted summing, biases the ensemble mean upward (since negative draws get clipped at the
  per-base level rather than allowed to cancel against positive draws in other bases). For disease counts the clip is right, but it should happen on the
  ensemble sample, not per base: ens_samp = np.maximum(Σ_i w_i (base_i + ε_i), 0).
  - No-bootstrap deterministic path emits a Samples with a single sample_0 column (_predictor.py:96-105). Downstream CRPS of a one-sample "distribution" is
  mathematically MAE. Either skip probabilistic metrics in this case or flag in the report so readers don't compare a single-sample CRPS against bootstrap-CRPS
   as if they're commensurable.
  - if s <= 0: w = uniform fallback in the predictor (_predictor.py:73-77). If you reach this branch the meta-model thinks every base is useless. Log a warning
   — it's almost always a sign of bad inner-val data.

behdadnikkhah25:
I’ve created a new draft pull request that compares against the master branch of my fork, rather than the stacking_new branch I used in previous iterations. Please follow the new draft PR, as it contains my latest code and addresses the feedback you gave me in the last pull request iteration.

PR #333, opened 2026-05-06
fix: return 200 with rejection summary on fully-rejected dry-run backtest (CLIM-582)

mortenoh:
## Summary

- `POST /v1/analytics/create-backtest-with-data/?dryRun=true` returned HTTP 400 when `provided_data` was empty or every region was rejected for missing covariate values. The modeling app expects dry-run to surface validation errors as HTTP 200 with an `ImportSummaryResponse` so the UI can render rejected regions uniformly.
- In dry-run mode, catch the 400 raised by `_read_dataset` / `_validate_full_dataset` and convert it to `ImportSummaryResponse(id=None, imported_count=0, rejected=...)`. Rejections are recovered from `exc.detail["rejected"]` (already serialised by `_validate_full_dataset`); the empty-data case yields `rejected=[]`.
- Non-dry-run behaviour is unchanged.

## Notes

- The recovery path couples to the dict shape produced by `_validate_full_dataset` (`detail["rejected"]`). A comment marks this. If that helper changes its detail format, this branch must be updated.
- Only 400 status codes are converted; any other error still propagates.
- `make_dataset` and the prediction endpoints are intentionally unchanged.

## Test plan

- [x] `uv run pytest tests/integration/rest_api/test_db_endpoints.py::test_backtest_with_empty_provided_data tests/integration/rest_api/test_db_endpoints.py::test_backtest_with_empty_provided_data_dry_run tests/integration/rest_api/test_db_endpoints.py::test_backtest_with_all_regions_rejected_dry_run`
- [x] `make lint`
- [x] `make test` (746 passed)

### DIFF
--- a/chap_core/rest_api/v1/routers/analytics.py
+++ b/chap_core/rest_api/v1/routers/analytics.py
@@ -583,8 +583,18 @@ async def create_backtest_with_data(
     database_url: str = Depends(get_database_url),
     worker_settings=Depends(get_settings),
 ):
-    feature_names, provided_data_processed = _read_dataset(request)
-    provided_data_processed, rejections = _validate_full_dataset(feature_names, provided_data_processed)
+    try:
+        feature_names, provided_data_processed = _read_dataset(request)
+        provided_data_processed, rejections = _validate_full_dataset(feature_names, provided_data_processed)
+    except HTTPException as exc:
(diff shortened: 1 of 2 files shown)


knutdrand:
I think @eirikhaugstulen  should check if this behavior is what FE expects

PR #335, opened 2026-05-07
Refactor/clim 590 clean session wrapper

ferenckata:
Remove unused methods `get_backtest_with_truth` and `seed_with_session_wrapper` , and remove `predictor/published_model.py` that never seem to be used either.

Refactored `add_model_template_from_yaml_config` and `add_model_template` so that the logic is cleaner and not repeated.  `add_model_template` is used in only one place, that got updated to the new signature.


knutdrand:
The helper functions should be private

chap_core/database/database.py
```diff
@@ -90,64 +90,53 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+        return existing_template
+
+    def return_model_template_id(self, model_name: str, existing_template: ModelTemplateDB) -> int:
+        logger.info(f"Model template with name {model_name} already exists. Returning existing id")
+        return cast("int", existing_template.id)
+
+    def update_model_template(self, existing_template: ModelTemplateDB, new_model_template: ModelTemplateDB) -> int:
+        logger.info(f"Model template with name {new_model_template.name} already exists. Updating it")
+        # Update the existing template with new data except id
+        data = new_model_template.model_dump(exclude={"id"})
```
knutdrand: The new `update_model_template` builds `data` from `new_model_template.model_dump(exclude={"id"})`, which materializes every column on `ModelTemplateDB` with its default — including DB-only columns that are not in `ModelTemplateConfigV2`. The notable casualty is `uses_chapkit`: it gets clobbered to `False` on every yaml-config re-registration. In the chapkit sync path it's masked because `crud.py:_sync_chapkit_services` re-sets `template.uses_chapkit = True` immediately after, but if that explicit re-set is ever removed or the yaml-update is called from a path that doesn't follow up with a re-set, the chapkit flag is silently lost. The new test `test_yaml_update_preserves_uses_chapkit` (added in aad8c01a) currently fails for exactly this reason.

Suggested fix — pass `exclude_unset=True` so the dump only emits keys that were actually present in the yaml-derived dict that constructed `new_model_template`:

```suggestion
        data = new_model_template.model_dump(exclude={"id"}, exclude_unset=True)
```

This restores the old contract ("update only yaml-supplied fields, leave DB-only columns alone"), keeps the redundant `existing_template.archived = False` line working as the explicit unarchive, and makes the new test pass. Verified locally against `test_yaml_update_preserves_uses_chapkit`, `test_add_model_template_unarchives_existing`, and `test_add_model_template_from_yaml_config`.

PR #336, opened 2026-05-07
fix: round disease_cases to int when updating historic data in Extend…

audunlas:
When ExtendedPredictor chains predictions iteratively, the mean of the sample columns (used as disease_cases for the next iteration) produces floats. Some models expect integer case counts, so this rounds the value before feeding it back as history.


knutdrand:
It should use median, not mean I think. Maybe put that in the same pr?

PR #347, opened 2026-05-11
feat: add counterfactual comparison plot for causal analysis

borissim:
## Summary

- Adds `plot_counterfactual` in `chap_core/assessment/causal_plot.py` that produces a per-location side-by-side Altair chart comparing original vs counterfactual predictions. Both panels for a given location share the same y-axis via Altair's native `resolve_scale(y="shared")` so case counts are directly comparable.
- Adds a `--plot` flag to `causal_cmd` that saves the comparison chart as an HTML file alongside the NetCDF output.
- Adds a `--cf-start-period` argument to `causal_cmd` so that counterfactual covariates can be introduced before `split_period` (i.e. as historical context during prediction), with original data used prior to that point.
- Exposes 6 years of historical context in both original and counterfactual `Evaluation` objects so the plot includes pre-split observations.
- Promotes `Evaluation._calculate_periods_from_years` and `Evaluation._extract_historical_observations` to public methods so they can be reused from `causal_cmd` without accessing private API.
- Adds a `y_domain` parameter to `EvaluationPlot.plot` so callers can enforce a fixed y-axis scale across multiple charts.

## Test plan

- [x] `tests/evaluation/test_causal_plot.py` — unit tests covering `plot_counterfactual`, shared y-domain, and single/multi-location rendering
- [x] `tests/test_causal_cmd.py` — tests for `cf_start_period` validation and the `--plot` flag
- [x] `make lint && make test` pass locally

### DIFF
--- a/chap_core/assessment/backtest_plots/evaluation_plot.py
+++ b/chap_core/assessment/backtest_plots/evaluation_plot.py
@@ -206,7 +206,7 @@ class EvaluationPlot(BacktestPlotBase):
             .mark_line()
             .encode(
                 x="time_period:T",
-                y=alt.Y("q_50:Q", alt.Scale(zero=False))),
+                y=alt.Y("q_50:Q", alt.Scale(zero=False)),
             )
         )
 
--- a/chap_core/assessment/backtest_plots/evaluation_plot.py
+++ b/chap_core/assessment/backtest_plots/evaluation_plot.py
@@ -206,7 +206,7 @@ class EvaluationPlot(BacktestPlotBase):
             .mark_line()
             .encode(
                 x="time_period:T",
-                y=alt.Y("q_50:Q", alt.Scale(zero=False)),
+                y=alt.Y("q_50:Q", scale=alt.Scale(zero=False)),
             )
         )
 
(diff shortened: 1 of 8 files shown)


knutdrand (commented):
Check if there is a flag when faceting charts in Altair to keep y-axis shares on rows. Ideally we don't want to change the interface of EvaluationPlot with the new kw arg

borissim:
@knutdrand ready for another review.
Remove the y domain calculation and replaced with an altair flag.
W.r.t. the Evaluation interface, I will create a task for it when I come up with the proper design.

knutdrand:
Two things from the review worth addressing before merge:

**Stray import in `tests/test_causal_cmd.py`**

```python
from xarray import plot
```

This is unused — looks like an editor autoimport triggered by the `plot=True` kwarg added below. Should be removed (lint may catch it too).

**`to_flat()` called once per location in `causal_plot.py`**

In `_chart_for_location`, `evaluation.to_flat()` runs every time the function is called, but it returns the same data on every call. With N locations and 2 evaluations that's 2N redundant `to_flat()` invocations plus 2N `pd.DataFrame(...)` wraps. Suggestion: hoist the `flat = evaluation.to_flat()` and the obs/forecasts/historical DataFrame construction out into `plot_counterfactual` (once per evaluation), then pass already-filtered per-location frames into `_chart_for_location`.

borissim:
@knutdrand after some rebase kung fu this should be ready to merge. There was some leftovers of the y-domain scaling thingy in evaluation_plot.py that made it harder to determine if the merge conflict resolution went fine. It is also fixed now next to your suggestions, good catch btw.

knutdrand (approved):

PR #351, opened 2026-05-18
fix: warn on unused CSV columns when running a model

borissim:
## Summary

Closes CLIM-709.

When a user runs `chap eval` or `chap causal` with a dataset that contains covariate columns not required by the model, CHAP silently accepted all columns without feedback. This made it easy to miss misconfigured datasets — for example, running a model with no required covariates against a dataset with temperature and rainfall data, with none of it being used.

## Changes

- **`chap_core/services/dataset_validation.py`** — Added `_check_unused_covariates()`, which compares dataset field names against the model's `required_covariates`. Each column present in the CSV but not referenced by the model produces a `warning`-level `ValidationIssue`. Wired into the existing `validate_dataset()` pipeline (also triggered by `chap validate --model-name`). A new `additional_continuous_covariates` parameter allows callers to pass user-configured extra covariates so they are not flagged.

- **`chap_core/cli_endpoints/_common.py`** — Added `warn_unused_covariates()` helper that calls `validate_dataset()` and routes issues to the logger without aborting execution.

- **`chap_core/cli_endpoints/evaluate.py`** / **`causal.py`** — Both now call `warn_unused_covariates()` after the estimator is instantiated, forwarding `ModelConfiguration.additional_continuous_covariates` when a config YAML was provided.

- **`tests/test_validate_cli.py`** — Three new tests covering: unused covariate warning, the `allow_free_additional_continuous_covariates=True` + no explicit columns case (still warns), and the `allow_free + explicit columns` case (suppresses warning for configured columns).

## Behaviour details

| Scenario | Result |
|---|---|
| Model requires `["rainfall"]`, dataset has `rainfall`, `temperature` | Warning for `temperature` |
| Model requires `[]`, `allow_free=False` | Warning for every covariate column |
| Model requires `[]`, `allow_free=True`, no `additional_continuous_covariates` | Warning for every covariate column |
| Model requires `["rainfall"]`, `allow_free=True`, `additional_continuous_covariates=["temperature"]` | No warnings |

Warnings are emitted via `logger.warning` and do not abort model execution.


borissim:
@knutdrand this one was spawned from feat/clim-696-counterfactual-plot, it needs to be rebased on master when #347 is merged, before merging.

knutdrand:
On `chap_core/cli_endpoints/_common.py` (the `warn_unused_covariates` helper):

This imports `_check_unused_covariates` directly from `chap_core.services.dataset_validation` — i.e. reaches across modules into a private (underscore-prefixed) symbol. Project CLAUDE.md rule 13 ("Never access private variables") forbids this.

Since two modules now depend on this check, it is effectively public. Suggest renaming `_check_unused_covariates` → `check_unused_covariates` in `dataset_validation.py` and updating both the import in `_common.py` and the docstring line "Calls `_check_unused_covariates` directly".

chap_core/cli_endpoints/evaluate.py
```diff
@@ -215,6 +216,7 @@ def _run_eval(
         estimator: ExternalModel | HpoModel | ExtendedPredictor
         if estimator_options.mode == EstimatorMode.NORMAL:
             estimator, configuration = get_estimator(template, model_configuration_yaml)
+            warn_unused_covariates(dataset, template.model_template_config, configuration)
```
mortenoh: Is this not relevant for HPO mode?

mortenoh (approved):

PR #353, opened 2026-05-18
test: smoke tests for backtest plot faceting API (CLIM-548)

knutdrand:
## Overview
This PR resolves performance bottlenecks and frontend layout constraints encountered when rendering custom backtest visualizations.

Historically, generating a single, monolithic Vega specification for an entire backtest caused browser-side rendering timeouts and bloated network payloads for large datasets. Additionally, hardcoded plotting structures stripped layout control away from the user interface.

To address this, we have decoupled visual generation from layout logic by implementing a decentralized, registry-backed plugin architecture. Visualizations can now explicitly define their sliceable metadata, allowing the frontend to programmatically request tailored coordinates, isolated single subplots, or a matrix of subplots.

### 1. Core Architecture & Plugin Registry
BacktestPlotBase: Introduced a minimal, clean abstract base class for all visualizations, ensuring backward compatibility for non-faceted layouts (e.g., summary dashboards or sample bias plots).

@backtest_plot Decorator: Implemented an automated registration system for scanning and tracking available visualization classes across the backend framework via a centralized dictionary registry.

FacetDimension: A dedicated dataclass introduced to capture visual configuration metadata, cleanly separating internal dataframe column mappings from user-facing layout names.

### 2. The FacetedBacktestPlot Refactor Split
Following internal design feedback, faceting logic was isolated away from the root base class into a dedicated intermediate subclass. This guarantees that non-faceted plots aren't forced to implement layout plumbing they don't use.

Pipeline Pattern (_preprocess + _plot): Subclasses are now driven by a strict data preparation vs. chart definition paradigm.

Deterministic Coordinate Slicing: Out-of-the-box functions calculate unique, sorted coordinate boundaries directly against the preprocessed dataframes.

True Subplots: Fixed a design flaw where subplots were actually degenerate 1x1 facet charts. The updated get_subplot() mechanism applies targeted boolean masks to the preprocessed data and directly outputs clean, lightweight, unfaceted Altair base instances.

### 3. Optimized EvaluationPlot Implementation
Vectorized Data Pipelines: Rewrote data aggregation routes. Heavy loop-driven copying was replaced with memory-optimized Pandas logic (groupby().quantile() transformations and cross-joins) to map historical contexts over dynamic split boundaries.

Multi-layered Altair Visuals: Generates layered charts combining standard line tracks for factual observations, median predictions, and stacked mark_errorband segments displaying 10–90% and 25–75% forecasting uncertainty horizons.

#### GET /backtest-plots/{visualization_name}/{backtest_id}/facet-coords
Returns valid dimension coordinate pairings available for building UI layout selectors.

#### POST /backtest-plots/{visualization_name}/{backtest_id}/subplot
Generates a single, lightweight Vega-compliant schema specification for an isolated coordinate slice.

#### GET /backtest-plots/{visualization_name}/{backtest_id}/subplots
Generates a flat checklist matrix mapping every valid combination variant against its respective standalone chart spec.

## Architectural Impact

By removing hardcoded dimensional knowledge (like location or split_period) from the base plotting plumbing, the system can seamlessly scale to any new custom facet dimensions (e.g., horizon_distance) defined by future plugins without breaking the core contract.

### DIFF
--- a/chap_core/assessment/backtest_plots/__init__.py
+++ b/chap_core/assessment/backtest_plots/__init__.py
@@ -4,47 +4,6 @@ Backtest visualization plugin system.
 This module provides a clean interface for creating custom backtest visualizations.
 Each visualization is a single class that receives flat pandas DataFrames and returns
 an Altair chart.
-
-Example usage:
-    from chap_core.assessment.backtest_plots import backtest_plot, BacktestPlotBase
-    import pandas as pd
-    import altair as alt
-
-    @backtest_plot(
-        plot_id="my_custom_plot",
-        name="My Custom Plot",
-        description="A custom visualization showing forecast accuracy."
-    )
-    class MyCustomPlot(BacktestPlotBase):
-        def plot(
-            self,
-            observations: pd.DataFrame,
-            forecasts: pd.DataFrame,
-            historical_observations: pd.DataFrame | None = None
-        ) -> alt.Chart:
-            # observations has columns: location, time_period, disease_cases
-            # forecasts has columns: location, time_period, horizon_distance, sample, forecast
-            # historical_observations (optional): location, time_period, disease_cases
-            ...
-            return chart
-
-Data schemas:
-    observations (pd.DataFrame):
-        - location: str - Location identifier
-        - time_period: str - Time period (e.g., "2024-01" or "202401")
-        - disease_cases: float - Observed disease cases
-
-    forecasts (pd.DataFrame):
-        - location: str - Location identifier
-        - time_period: str - Time period being forecasted
-        - horizon_distance: int - How many periods ahead this forecast is
-        - sample: int - Sample index (for probabilistic forecasts)
-        - forecast: float - Forecasted value
-
-    historical_observations (pd.DataFrame, optional):
-        - location: str - Location identifier
-        - time_period: str - Time period
-        - disease_cases: float - Historical observed disease cases
 """
 
 from __future__ import annotations
@@ -73,20 +32,13 @@ class BacktestPlotBase(ABC):
 
     Subclasses must implement the `plot` method which receives flat DataFrames
     and returns an Altair chart.
-
-    Attributes:
-        id: Unique identifier for the plot (set by decorator)
-        name: Display name for the plot (set by decorator)
-        description: Description of what the plot shows (set by decorator)
-        needs_historical: Whether this plot needs historical observations (set by decorator)
     """
 
     id: str = ""
     name: str = ""
     description: str = ""
     needs_historical: bool = False
-
-    facet_dimensions: list[str] = []
+    facet_dimensions: list[str] = []  
 
     @abstractmethod
     def plot(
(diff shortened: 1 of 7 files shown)


knutdrand:
## Proposed refactor before merge: split faceting into a `FacetedBacktestPlot` subclass

The current implementation works, but `BacktestPlotBase` ends up with hard-coded knowledge of `location` and `split_period` and the faceting methods sit on a class that's also used by plots that don't facet. Three concrete issues this causes today:

1. `get_full_plot` hard-codes `"location" → row`, `"split_period" → column`. Any subclass declaring a different `facet_dimensions` (e.g. `["horizon_distance"]`) silently drops the facet.
2. `get_subplot` filters dataframes by `coord in df.columns`. `split_period` is a *derived* column (computed inside `EvaluationPlot.plot` via `_infer_split_periods`), so passing `{"split_period": ...}` silently does no filtering on forecasts. The smoke test happens to pass because `next(iter(coords))` returns `"location"` first.
3. `get_subplot` calls `get_full_plot`, which re-applies `.facet(...)`, so a "subplot" is actually a degenerate 1x1 `FacetChart` wrapper.

### Proposed shape

Introduce an intermediate subclass that owns the faceting workflow and forces every faceted plot into a preprocessing + plotting split:

```python
class FacetedBacktestPlot(BacktestPlotBase):
    facet_dimensions: list[str] = []   # altair shorthand, e.g. ["location:N", "split_period:O"]

    @abstractmethod
    def _preprocess(self, observations, forecasts, historical_observations=None) -> pd.DataFrame:
        """Combine inputs into a single dataframe carrying every column the plot consumes,
        including derived ones like `split_period`."""

    @abstractmethod
    def _plot(self, df: pd.DataFrame) -> ChartType:
        """Render the unfaceted base chart from the preprocessed dataframe."""

    # The faceting API then becomes generic and lives entirely on the base:
    def plot(self, observations, forecasts, historical_observations=None):
        return self._plot(self._preprocess(observations, forecasts, historical_observations))

    def facet_coords(self, observations, forecasts, historical_observations=None):
        df = self._preprocess(observations, forecasts, historical_observations)
        return {dim.split(":", 1)[0]: sorted(df[dim.split(":", 1)[0]].dropna().unique().tolist())
                for dim in self.facet_dimensions}

    def get_subplot(self, observations, forecasts, coords, historical_observations=None):
        df = self._preprocess(observations, forecasts, historical_observations)
        for col, value in coords.items():
            df = df[df[col] == value]
        return self._plot(df)

    def get_full_plot(self, observations, forecasts, historical_observations=None):
        chart = self.plot(observations, forecasts, historical_observations)
        if not self.facet_dimensions:
            return chart
        kwargs = dict(zip(["column", "row"], self.facet_dimensions, strict=False))
        return chart.facet(**kwargs).resolve_scale(y="independent")
```

### What this buys

- **`BacktestPlotBase` stops knowing about specific facet dimensions.** Faceting code references `self.facet_dimensions` and nothing else; the per-dim row/column mapping and the split_period derivation both disappear from the base.
- **Type markers on `facet_dimensions` (`"location:N"`)** make `.facet(column=...)` work generically — altair accepts the shorthand directly, so no per-dim if-ladder.
- **Filtering for subplots becomes correct by construction.** `_preprocess` is responsible for putting `split_period` (or any derived facet column) on the dataframe. `get_subplot` filters by column equality on that dataframe; there is no "column missing → silently no-op" branch.
- **Subplots are no longer 1x1 facet wrappers.** `get_subplot` calls `_plot(filtered_df)` directly; `.facet(...)` is only applied by `get_full_plot`.
- **Plots that don't facet (`SampleBiasPlot`, `MetricsDashboard`, ...) stay on plain `BacktestPlotBase`** until they opt in. No forced migration.

### Migration plan

Two plots in this PR, the rest follow incrementally:

- `EvaluationPlot` — `facet_dimensions = ["split_period:O", "location:N"]` (order chosen so column=split_period, row=location matches today's layout). `_preprocess` returns the `combined_data` dataframe currently built inline in `plot()`. `_plot` returns the unfaceted layered chart.
- `PredictedVsActualPlot` — `facet_dimensions = ["horizon_distance:O"]`. Different from EvaluationPlot's dims on purpose: it exercises the generic path in the base class and proves nothing in `FacetedBacktestPlot` is `EvaluationPlot`-specific. `_preprocess` returns the merged + log1p'd dataframe; `_plot` returns the scatter+identity layer without `.facet(...)`.

### Tests

Smoke tests in `tests/evaluation/test_backtest_plot_faceting.py` already document this shape (parametrized over both plots, `xfail(strict=True)` on the refactor-specific structural assertions; the behavioral assertions run live as regression checks). REST endpoint smoke tests in `tests/integration/rest_api/test_backtest_plot_faceting_api.py` cover the corresponding HTTP surface (also `xfail(strict=True)`).

### Lint cleanup riding along

`make lint` currently exits non-zero on this branch:

- `C414` `sorted(list(...))` in `facet_coords` (`__init__.py:143`)
- `B905` `zip()` without `strict=` in `get_subplots` (`__init__.py:183`)

plus 15 auto-fixable issues (trailing whitespace on lines 88/89/121/181/193/195, missing trailing newlines on both files, unused top-level `import itertools` on line 54 — duplicated inline on line 177). Worth fixing in the same refactor pass.

chap_core/assessment/backtest_plots/__init__.py
```diff
@@ -72,45 +32,144 @@ class BacktestPlotBase(ABC):
...
+        return self.plot(observations, forecasts, historical_observations)
+
+
+class FacetedBacktestPlot(BacktestPlotBase):
+    """
+    Subclass managing generic programmatic grid faceting and subplots.
+    Separates concerns into a definitive data preprocessing and visual layout pipeline.
+    """
+
+    facet_dimensions: list[str] = []  # Altair shorthand, e.g., ["split_period:O", "location:N"]
```
knutdrand: I think this will also need to keep display name, since it is used for showing the dimension in predicted_vs_actual_plot

chap_core/assessment/backtest_plots/__init__.py
```diff
@@ -72,45 +32,144 @@ class BacktestPlotBase(ABC):
...
+
+        # Handle native Altair types cleanly without dictionary unpacking unpacks
+        if col and row:
+            faceted_chart = chart.facet(column=col, row=row)
+        elif col:
+            faceted_chart = chart.facet(column=col)
+        else:
+            faceted_chart = chart
+
+        return faceted_chart.resolve_scale(y="independent")  # type: ignore[no-any-return]
```
knutdrand: We're forcing every plot to have y independent here. The individual plot implementations need to be able to specify this. I think for instance the predicted_vs_actual_plot needs it to be shared

knutdrand:
Take away xfail from any tests that should pass, and run the test suite, I think there will be some errors, i.e. 1. Tests don't match the new dataclass shape (will fail). tests/evaluation/test_backtest_plot_faceting.py
   still asserts isinstance(entry, str) and applies FACET_DIM_RE to facet_dimensions entries. But entries
  are now FacetDimension instances, not strings. test_facet_dimensions_use_type_marker_shorthand and
  test_every_faceted_plot_in_registry_passes_shape_checks will both fail. Either update the tests to
  inspect entry.field_name, or pivot the assertions onto the dataclass type.
  2. Typo: :0 instead of :O in PredictedVsActualPlot.
  facet_dimensions = [FacetDimension(field_name="horizon_distance:0", display_name="Horizon Distance")]
  2. That's the digit zero, not letter O. Altair won't recognize it as the ordinal encoding type — the
  facet column likely degrades silently. Should be "horizon_distance:O"

Once tests pass, update PR description and mark as ready to review

yambanso (approved):
seems good

PR #359, opened 2026-05-20
feat: add causal build-counterfactual subcommand

borissim:
## Summary

Adds a `build-counterfactual` subcommand under `chap causal` (CLIM-697). It reads a CHAP-compatible CSV, applies user-supplied arithmetic transformations to one or more columns, and writes the result to a new CSV file. This makes it straightforward to produce counterfactual datasets for use with `chap causal`.

## New command

```
chap causal build-counterfactual <dataset.csv> <col=expr> [<col=expr> ...] [options]
```

**Options:**
- `--start-time-period` / `--end-time-period` — restrict transformations to a time-period range (inclusive on both ends)
- `--output-csv` — override the default output path (default: input filename with `_cf` suffix, e.g. `data.csv` → `data_cf.csv`)

**Expression grammar** — each expression is an arithmetic formula over `x` (the original cell value):
- Operators: `+`, `-`, `*`, `/`, `**`
- Functions: `abs(x)`, `round(x)`
- Nesting: `abs(x*0.1-5)`, `round(x+0.5)`
- Missing values (NaN) are left unchanged regardless of the expression

**Examples:**
```
chap causal build-counterfactual data.csv rainfall=x*0.01 temperature=x-30
chap causal build-counterfactual data.csv rainfall=abs(x) --start-time-period 2022-06
chap causal build-counterfactual data.csv cases=round(x*1.1) --output-csv data_cf.csv
```

## Changes

- `chap_core/cli_endpoints/build_counterfactual.py` — new module implementing:
  - `_validate_expression` — AST-based safety check; only `x`, numeric literals, arithmetic operators, and `abs`/`round` are allowed
  - `_parse_transformations` — parses `['col=expr', ...]` into `[('col', 'expr'), ...]`
  - `_apply_transformation` — applies an expression to a Series, leaving NaN values untouched
  - `build_counterfactual_cmd` — the cyclopts CLI entry point with full validation and logging
- `chap_core/cli_endpoints/causal.py` — `register_commands` now mounts `build-counterfactual` as a subcommand of the `causal` app
- `tests/test_build_counterfactual.py` — 46 tests covering all three private helpers and the command end-to-end (type coercion warning, time-period filtering, custom output path, validation errors)
- `tests/conftest.py` — adds shared `make_test_df` fixture (factory for minimal CHAP-compatible DataFrames) used by both the new tests and the existing causal-cmd tests
- `tests/test_causal_cmd.py` — migrated from a local `_make_df` helper to the shared `make_test_df` fixture

## Test plan

- [ ] `uv run pytest tests/test_build_counterfactual.py` — all 46 tests pass
- [ ] `uv run pytest tests/test_causal_cmd.py` — existing causal tests unaffected
- [ ] `uv run chap causal build-counterfactual --help` — subcommand is listed and help text renders correctly


knutdrand:
A few follow-ups from the review:

- **Test function names** — `testvalidate_expression_*`, `testparse_transformations_*`, and `testapply_transformation_*` are missing the underscore after `test`. Pytest still discovers them, but every other test in the repo uses `test_<name>`. Worth renaming for consistency.

- **`eval` re-parses per cell** — `FeatureTransformations.apply_transformation` does `series[mask].apply(lambda x: eval(expr, namespace, {\"x\": x}))`, which re-parses `expr` for every value. Compiling once is a free speedup:
  ```python
  code = compile(expr, \"<expr>\", \"eval\")
  result[mask] = series[mask].apply(lambda x: eval(code, namespace, {\"x\": x}))
  ```

- **`apply_transformation` doesn't enforce validation** — It's a public static method that calls `eval` on the raw `expr`. The CLI happens to call `validate_expression` first, but a future caller could skip it. Either fold validation into `apply_transformation`, or document the precondition explicitly in its docstring.

borissim:
@knutdrand all good stuff, ready for another round.

knutdrand (approved):

PR #362, opened 2026-05-20
feat: filter regions with no disease cases in chap eval

knutdrand:
## Summary
- The REST API filters out regions whose entire training-period `disease_cases` is NaN before running the backtest (via `validate_and_filter_dataset_for_evaluation` in `chap_core/rest_api/db_worker_functions.py`).
- `chap eval` (CLI) was not applying this filter, so the model received locations with no usable target data. This caused divergent behaviour between the two entry points.
- Call the same filter in `eval_cmd` right after loading the dataset, using `backtest_params` for `n_periods` / `n_splits` / `stride`.
- Also fixes a latent bug in `validate_and_filter_dataset_for_evaluation`: `rejected` was declared but never populated, so the "Rejected regions" warning always logged an empty list and fired unconditionally. It is now populated correctly and only warns when regions are actually dropped.

## Merge with master
- Resolved a conflict in `tests/test_cli.py` where master and this branch each added new `eval_cmd` tests at the same location; both sets are kept.
- Master's new dispatch/tracking tests mock the dataset as a `MagicMock`. The new filter iterates `dataset.items()`, which collapses a `MagicMock` to an empty `DataSet` and broke those tests. The shared `_patched_eval_chain` helper takes `patch_filter=True` (default) to stub the filter as a passthrough for tests not concerned with filtering; the filter test passes `patch_filter=False` to exercise the real filter.

## Test plan
- [x] `make lint` — passes
- [x] `test_eval_cmd_drops_regions_with_no_disease_cases` — builds a dataset with one valid region and one all-NaN region, runs `eval_cmd`, and asserts the dataset forwarded to `Evaluation.create` no longer contains the all-NaN region (proves `eval_cmd` wires the filter in).
- [x] `test_validate_and_filter_drops_region_with_all_nan_training_data` — exercises the filter directly on its returned `DataSet` for the rejection semantics.

## Note on the earlier test-all failure
The first version of the region-filter test asserted on `caplog` records for the "Rejected regions" warning. That warning only reaches `caplog` via root-logger propagation, and in the `test-all` workflow earlier integration tests mutate global logging state, so the warning was not captured and the assertion failed (`assert 0 == 1`) even though the filter worked. The log-capture assertions were dropped in favour of asserting the filter contract directly (the returned `DataSet`), which is deterministic and independent of suite ordering.


Halvardgithub:
Should state which regions are removed, current: 
 [WARNING] Rejected regions: [] due to missing target values for the whole training period [chap_core.rest_api.db_worker_functions]
2026-05-20T15:26:21 [INFO   ] Remaining regions: ['LA-AT', 'LA-BK', 'LA-BL', 'LA-CH', 'LA-HO', 'LA-KH', 'LA-LM', 'LA-LP', 'LA-OU', 'LA-PH', 'LA-SL', 'LA-SV', 'LA-VT', 'LA-XA', 'LA-XE', 'LA-XI', 'LA-XN'] with 17 entries [chap_core.rest_api.db_worker_functions]

knutdrand:
Good catch — that was a pre-existing bug in `validate_and_filter_dataset_for_evaluation`: the `rejected` list was initialised empty and never populated, so the warning always read `Rejected regions: []` even when regions were actually dropped. Verified against `chap_LAO_admin1_monthly-3.csv`: 18 input locations vs 17 in the `Remaining regions` line — `LA-VI` was silently filtered out.

Fixed in 73654653: build the `rejected` list while iterating, and only emit the warning when something was actually dropped. The CLI test now also asserts the dropped region name appears in the warning record.

mortenoh (approved):

PR #365, opened 2026-05-21
feat: select HPO searcher via estimator options

zlilu:
## Summary

Lets `chap eval` choose the HPO searcher (grid / random / tpe) through `--estimator-options.searcher`, instead of the searcher being hardcoded. Also makes the HPO `metric` optional with a sensible default.

## Changes

- **`api_types.py`**: add `SearcherType` enum (`grid`, `random`, `tpe`) and a `searcher: SearcherType | None` field on `EstimatorOptions`. The `metric` field now defaults to `None` (resolved to a default downstream) rather than `"rmse"`.
- **`cli_endpoints/evaluate.py`**: HPO mode forwards `estimator_options.searcher` to `get_hpo_estimator` and updates the `--estimator-options` help text.
- **`cli_endpoints/_common.py`**: `get_hpo_estimator` takes `searcher_inp: SearcherType | None` and `metric: str | None`, maps the searcher type to the matching `Searcher` instance, and raises `ValueError` on an unknown value instead of silently defaulting.
- **`hpo/hpoModel.py`**: `searcher` is now optional; when `None`, `HpoModel` defaults to `RandomSearcher(DEFAULT_SEARCH_TRIALS)`. Constructor argument order updated accordingly.
- **`hpo/objective.py`**: `metric` is optional and resolves to `"rmse"` when not provided.
- **`hpo/searcher.py`**: add `DEFAULT_SEARCH_TRIALS`, validate that `GridSearcher` only accepts list-based search spaces, and make `TPESearcher`'s `max_trials` an explicit first argument.
- **`hpo/base.py`**: remove a stray debug `print` and a block of dead commented-out dataclasses.

## Try it

The search space comes from the model template's `hpo_search_space`. `chap-models/minimal_template_example` defines one (`alpha`), so this runs end to end:

```bash
uv run chap eval \
    --model-name https://github.com/chap-models/minimal_template_example \
    --dataset-csv example_data/laos_subset.csv \
    --output-file /tmp/chap/min_hpo_eval.nc \
    --backtest-params.n-splits 2 \
    --backtest-params.n-periods 1 \
    --estimator-options.mode=hpo \
    --estimator-options.searcher=grid
```

`--estimator-options.searcher` accepts `grid`, `random`, or `tpe`. With `grid` above, the searcher sweeps every `alpha` in the template's search space before exporting the evaluation NetCDF.

## Tests

- `tests/hpo/test_hpo_estimator.py` (new): covers the `SearcherType` to searcher dispatch, the `None` passthrough, and the unknown-value path raising.

## Test plan

- [x] `make lint`
- [x] `tests/hpo/` and `tests/test_cli.py` pass
- [x] Ran the command above end to end (grid sweep over the template `hpo_search_space`, NetCDF written)


mortenoh:
Typos in the searcher name silently fall through to the default

  File: chap_core/cli_endpoints/_common.py:280-286

  The dispatch is a chain of if/elif with no else. If searcher_inp is anything other than "grid", "random", or "tpe", the local searcher stays None, and HpoModel's constructor then
  silently defaults it to RandomSearcher(3). The CLI is shielded from this because pydantic validates the enum at parse time — but anyone calling get_hpo_estimator() from Python
  (tests, notebooks, future programmatic callers) gets the silent-default behavior. I reproduced this with searcher_inp="bogus" returning a working RandomSearcher with no warning.

  Fix: Add an explicit else: raise ValueError(...). Even better, tighten the type to SearcherType | None so mypy/pyright catches the typo statically.

mortenoh:
Follow-up review round. The earlier comment had two parts; one was already handled, one wasn't:

- Type tightening to \`SearcherType | None\` was in place (catches typo strings statically). Good.
- The explicit \`else: raise\` was still missing, so an unhandled-but-valid searcher value (e.g. a future 4th \`SearcherType\` member whose \`elif\` is forgotten) would silently fall back to the default \`RandomSearcher\`.

Pushed 77bec92 to close it out:
- Added \`else: raise ValueError(f"Unknown searcher: {searcher_inp!r}")\` in \`get_hpo_estimator\`.
- Removed a leftover commented-out \`# assert estimator_options.metric is not None\` in \`evaluate.py\`.
- Added \`tests/hpo/test_hpo_estimator.py\` covering the grid/random/tpe dispatch, the \`None\` passthrough, and the unknown-value path now raising.

\`make lint\` clean, \`tests/hpo/\` + \`tests/test_cli.py\` green.

mortenoh (approved):

PR #387, opened 2026-05-27
Feat: Added region labels to model card output

bsmarkus:
Added region labels to the choropleth map plots based on features in GeoJSON file.

### DIFF
--- a/chap_core/cli_endpoints/generate_modelcard.py
+++ b/chap_core/cli_endpoints/generate_modelcard.py
@@ -206,23 +206,19 @@ def _save_evaluation_plots(evaluation: Evaluation, output_dir: Path, geojson_pat
 
         label_points = [
             {
-                "region_name": (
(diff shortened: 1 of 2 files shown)


bsmarkus:
@knutdrand Can you take a look at this?

mortenoh (changes requested):
Most of our geojson will not have a calculated `bbox` property so we need to include a bbox fallback (generate it ourself, look at `chap_core.geoutils.feature_bbox(feature)`). Aslo in the naming fallback chain, you should include `name` also as this is more common

mortenoh (changes requested):
Add a test for _label_points_from_geojson_features covering: a feature without bbox, name precedence (name > ADM1_EN > id), and malformed input being skipped — reuse the dummy_geojson fixture.

mortenoh (approved):

PR #401, opened 2026-06-02
Feat/clim 548 refactor backtest plots filters

yambanso:
Migrate EvaluationPlot and PredictedVsActualPlot to the faceted base and declare facet_dimensions.
Add REST support for facet coords, isolated subplot generation, and subplots listing.
Add smoke/regression tests for faceting API and shorthand validation.
Backwards-compatible: non-faceted plots keep working via existing plot/registry.


knutdrand (changes requested):
I think the horizon distance logic in the visualization endpoints should be removed. Rather make a unit test for PredictedvsActual plot that it returns the correct facet_coords and fix any bugs there

PR #408, opened 2026-06-03
feat: add n_retrain parameter to BacktestParams

knutdrand:
## Summary

Adds an `n_retrain` parameter to `BacktestParams` (CLIM-766) that controls how many times the model is retrained during backtesting.

- `n_retrain=1` (default): model is trained once and predictions run across all `n_splits` windows. Identical to the previous backtest.
- `n_retrain=2`: model is trained at the start and retrained once halfway through the splits.
- `n_retrain=k`: retrains at `k` evenly spaced split points, using the expanding window of data available at each point.

## Changes

- `api_types.py`: new `n_retrain` field (default 1, `gt=0`) plus a validator enforcing `n_retrain <= n_splits`.
- `prediction_evaluator.py`: `backtest()` gains `n_retrain`; the single up-front train is replaced by retraining inside the loop at the computed split indices (`_retrain_split_indices`). Split 0 trains on the dedicated `train_set` from `train_test_generator` (which carries the `_train_set` metadata), so the `n_retrain=1` default is byte-identical to the old code; later retrains train on that split's expanding historic window.
- `evaluation.py`: `Evaluation.create` passes `n_retrain` through.
- `_args.py`: documents `--backtest-params.n-retrain` in CLI help.

## Tests

- `test_prediction_evaluator.py`: asserts `train` is called once on `train_set` by default, and on `[train_set, historic2]` for `n_retrain=2` over 4 splits.
- `test_api_types.py`: default is 1; rejects `n_retrain > n_splits`.

`make lint` passes; related suites (evaluation, dataset-splitting, CLI args, REST validation) pass.

### DIFF
--- a/chap_core/api_types.py
+++ b/chap_core/api_types.py
@@ -16,7 +16,7 @@ from geojson_pydantic import (
     Point,
     Polygon,
 )
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, Field, model_validator
 
 from chap_core.database.base_tables import DBModel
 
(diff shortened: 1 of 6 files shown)


ivargr (approved):
Looks good.

PR #410, opened 2026-06-04
chore(deps): slim celery extra and drop unused diskcache

mortenoh:
## What

Dependency cleanup that removes two unused/test-only dependency chains and refreshes type stubs.

- **`celery[pytest]` -> `celery`**: the test suite uses `celery.contrib.pytest` (bundled in celery core, via `pytest_plugins = ("celery.contrib.pytest",)`), not the separate `pytest-celery` package. The `[pytest]` extra only pulled in `pytest-celery[all]` plus its `boto3`/`botocore`/`pycurl`/`debugpy` chain, none of which we use.
- **Removed `diskcache`**: it was declared as a direct dependency but is not imported anywhere in `chap_core/` or `tests/` (added during the original pip->uv migration). Caching in the codebase uses stdlib `functools.cache`.
- **Bumped `types-requests` stubs** and dropped a stale dev-dependency pin.

Dropping the `pytest-celery` extra also lets several transitive dependencies resolve to current versions.

## Net effect on the lock

Removes `pytest-celery`, `boto3`, `botocore`, `pycurl`, `debugpy`, `s3transfer`, `tenacity`, `python-memcached`, `pytest-docker-tools`, `diskcache`, and `types-urllib3`; advances transitive pins.

## Testing

- `make lint` clean
- `tests/test_celery.py` passes (confirms `celery.contrib.pytest` fixtures still work without the extra)
- `make test-all` (docker-compose integration flow) passes


ivargr (approved):
Good

PR #414, opened 2026-06-04
feat: make 'chap test' a real self-diagnostic

mortenoh:
## Summary

`chap test` was a placeholder that only logged a debug + info line (the debug line is invisible at the default level) and silently swallowed arbitrary `--<keyword>` args via `**base_kwargs`. It proved strictly less than `chap --version`. This turns it into a useful install smoke test.

## What it does now

```console
$ chap test
chap-core      2.0.0.dev1
Python         3.13.12  (/path/.venv/bin/python3)
platform       darwin

imports:
  numpy / pandas / xarray ... OK
  public API (data, fetch, ModelTemplateInterface) ... OK

All checks passed.
```

- Prints chap-core version, Python version/executable, and platform.
- Verifies the core runtime imports resolve: the scientific stack (`numpy`/`pandas`/`xarray`) and the lazy public API (`data`, `fetch`, `ModelTemplateInterface`).
- Exits non-zero (and prints which imports are missing) if any check fails, so it is usable in CI / post-install verification.
- Side-effect free: no network, Docker, or DB access. Also drops the `**base_kwargs` catch-all so the command no longer accepts arbitrary flags.

The check logic is a small `_health_checks()` helper returning `(label, ok, detail)` rows; the command formats them and sets the exit code.

## Tests

- `test_chap_test_reports_environment_and_passes` — asserts the version/import lines print and the command exits 0.
- `test_chap_test_exits_nonzero_when_a_core_import_fails` — monkeypatches a core import to fail and asserts the missing import is surfaced and the command exits 1.

## Verification

- `make lint` clean (0/0).
- New tests pass; `chap test` exits 0 locally.

### DIFF
--- a/chap_core/cli_endpoints/utils.py
+++ b/chap_core/cli_endpoints/utils.py
@@ -95,16 +95,64 @@ def write_open_api_spec(out_path: str):
         json.dump(schema, f, indent=4)
 
 
-def test(**base_kwargs):
-    """
-    Simple test-command to check that the chap command works
-    """
-    from chap_core.log_config import initialize_logging
+def _health_checks() -> list[tuple[str, bool, str]]:
+    """Probe that chap-core's core runtime imports resolve.
 
-    initialize_logging()
+    Returns one ``(label, ok, detail)`` row per check group; ``detail`` names
+    the offending imports when ``ok`` is False. Kept side-effect free (no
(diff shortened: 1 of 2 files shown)


ivargr (approved):
Good

PR #431, opened 2026-06-09
feat: prefilter backtest-plot facets from the database

knutdrand:
## Summary

Improves backtest-plot faceting performance by pushing coordinate filtering down to the database instead of flattening the entire backtest into memory and filtering in pandas.

Previously, rendering a single subplot (one location / split period / horizon) or just listing facet coordinates loaded every forecast sample of every location, period, and horizon, then filtered in pandas. This change serves the well-known facet dimensions directly from the database.

### `max_horizon_distance` on `Backtest` (first commit)
- New nullable `max_horizon_distance` column recording the largest 1-based horizon distance scored in a backtest.
- Populated at every creation site (`from_samples_with_truth`, `from_file`, naive simulator) via a shared `max_horizon_distance` helper built on the existing `horizon_diff`.
- Alembic migration for the column. Legacy rows stay `NULL` and fall back to deriving horizons from the forecasts.

### Database-backed facet dimensions (second commit)
- `DBFacetDimension` subclasses for `location`, `split_period`, and `horizon_distance`. Each knows how to (a) enumerate available coordinates with cheap queries and (b) translate a requested coordinate into SQL filters:
  - `location` -> `org_unit ==`
  - `split_period` -> `last_seen_period` match (`split = clean_time(last_seen - delta)`)
  - `horizon_distance` -> coordinates `1..max_horizon_distance`, filter via `tuple_(period, last_seen).in_(...)`
- `load_filtered_flat_data` loads only the matching forecast/observation rows and returns the same `FlatEvaluationData` shape, so plot code is unchanged.
- The `/facet-coords` and `/subplot` endpoints use the database-backed path when all of a plot's facet dimensions support it, and fall back to the full-scan path otherwise. The plot's in-memory filter stays as a correctness safety-net.

## Tests
- Helper and creation-site tests for `max_horizon_distance`.
- Equivalence tests: facet coordinates and filtered forecast row-sets match the previous full-scan path; split-period filters form a disjoint, complete partition of the forecasts; unknown coordinate returns no rows.
- Existing REST faceting integration tests pass through the rewired endpoints unchanged.

## Follow-ups (not in this PR)
- The `/subplots` batch endpoint still uses the full-load path.
- No `BacktestForecast` composite index yet; the new `WHERE` clauses (`org_unit`, `last_seen_period`, `period`) will table-scan until one is added.

### DIFF
--- a/alembic/versions/a7b8c9d0e1f2_add_max_horizon_distance_to_backtest.py
+++ b/alembic/versions/a7b8c9d0e1f2_add_max_horizon_distance_to_backtest.py
@@ -1,0 +1,33 @@
+"""add_max_horizon_distance_to_backtest
+
+Adds a nullable max_horizon_distance column to the backtest table. Existing rows
+are left NULL; consumers fall back to deriving horizons from the forecasts.
+
+Revision ID: a7b8c9d0e1f2
+Revises: f6a7b8c9d0e1
+Create Date: 2026-06-09
+
+"""
+
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+from alembic import op
+
+revision: str = "a7b8c9d0e1f2"
+down_revision: Union[str, Sequence[str], None] = "f6a7b8c9d0e1"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    """Add max_horizon_distance column to backtest table."""
+    op.add_column(
+        "backtest",
+        sa.Column("max_horizon_distance", sa.Integer(), nullable=True),
+    )
+
+
+def downgrade() -> None:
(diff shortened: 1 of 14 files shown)


mortenoh:
Pushed a small follow-up refactor (`f780e3d4`): the observation load in `load_filtered_flat_data` now routes through `DataSetManager.observations(...)` instead of a hand-rolled `select(Observation)`, so `Observation` table access lives in the data-access manager (consistent with the manager pattern). The now-unused `observation_filters` extension point was dropped from the DB facet dimensions; the `BacktestForecast` queries are left inline since they're the only such queries in the codebase. Behavior is identical — facet equivalence and REST faceting tests pass unchanged, and `make lint` (ruff/mypy/pyright) is clean.

mortenoh (approved):

PR #469, opened 2026-06-25
refactor(plotting): drop vegafusion and pyarrow, compile Vega with vl-convert

mortenoh:
## Summary

Removes `vegafusion` and `pyarrow` from the chart-rendering stack, cutting **~167 MB** from the install footprint (vegafusion ~52 MB + pyarrow ~115 MB), while the API continues to return the same compiled-Vega specs.

Resolves [CLIM-648](https://dhis2.atlassian.net/browse/CLIM-648).

## Background

`pyarrow` is **not used directly anywhere** in the codebase. It was pinned in `pyproject.toml` solely because `vegafusion`'s `pre_transform_spec` converts pandas DataFrames to Arrow via `pandas.__arrow_c_stream__`, which hard-requires pyarrow. The chain is:

```
altair  ->  compile_with_vegafusion  ->  vegafusion.pre_transform_spec
        ->  narwhals.__arrow_c_stream__  ->  pandas.__arrow_c_stream__  ->  pyarrow
```

`vegafusion` itself was enabled only as Altair's **data transformer** (`alt.data_transformers.enable("vegafusion")` in three plotting modules) to pre-aggregate chart data server-side. It is **not** required to produce Vega: `vl-convert-python` compiles Vega-Lite to Vega on its own.

## What changed

- Replace the three `alt.data_transformers.enable("vegafusion")` calls with `alt.data_transformers.disable_max_rows()` in `dataset_plot.py`, `backtest_plot.py`, and `yaml_plot.py`. The API keeps returning compiled Vega specs (`to_dict(format="vega")`), now compiled by vl-convert.
- Remove `vegafusion` and `pyarrow` from `pyproject.toml` (and `uv.lock`, which also drops their transitives `arro3-core` and `narwhals`).
- `vl-convert-python` stays — it does the Vega-Lite -> Vega compilation **and** renders the modelcard PNG/SVG exports.
- Reword three code comments that named vegafusion-specific behavior; the underlying defensive logic (empty-chart placeholder, ISO date coercion) is still valid for vl-convert and is unchanged.
- Pre-aggregate the two metric "sum" plots (`MetricByHorizonV2Sum`, `MetricByTimePeriodV2Sum`) in pandas before charting, mirroring the existing mean plots. They were the only charts still declaring an in-spec `sum(metric)` aggregate, which — without vegafusion — would have embedded the full raw metric frame for the browser to sum. On a 9-location backtest this drops the horizon-sum spec from 27 embedded rows to 3; the saving scales with location count.
- Drop a redundant in-spec `mean(metric)` aggregate in `MetricByTimePeriodV2Mean`: it already pre-aggregates via a pandas `groupby().mean()`, so the encoding now uses the plain `metric` field, matching `MetricByHorizonV2Mean` and the two sum plots. Every metric plot now feeds a pre-aggregated frame to Altair with no in-spec aggregate transform.

## Why this approach

The investigation considered three options for CLIM-648:

| Approach | Drops vegafusion + pyarrow | vl-convert | Frontend change | Responsive sizing |
|---|---|---|---|---|
| **A — keep compiled Vega (this PR)** | yes | stays | none | unchanged |
| B — switch API to `format="vega-lite"` | yes | stays (PNG only) | required rework | must re-implement; VConcat regresses |
| Replace altair entirely | n/a | n/a | full rewrite | n/a |

Both A and B drop the same two dependencies and both keep vl-convert (the modelcard PNG export needs it either way), so **B sheds no additional dependencies** — it only moves Vega compilation to the browser, at the cost of a cross-repo rework and a responsive-width regression for vertically-concatenated plots (the backend's `_fill_container_width` injects a Vega `containerSize()` signal that has no Vega-Lite equivalent). Replacing altair is a non-option: it is the only maintained Python -> Vega-Lite library and is the thin part of the stack, not the weight.

Approach A gets the full footprint win with the smallest, lowest-risk change and no frontend coordination.

## Trade-off

Without vegafusion's server-side pre-aggregation, charts embed their **full, un-pre-aggregated data**, so Vega specs are larger on the wire (e.g. an 8000-row aggregated bar chart measured ~300 KB vs ~2 KB pre-transformed). The 5000-row `MaxRowsError` cap is lifted via `disable_max_rows()`. For the current plot data volumes this is acceptable; if spec sizes become a concern later, the lighter follow-up is to pre-aggregate in pandas before charting rather than re-introduce vegafusion.

## Testing

- `make lint` — ruff, mypy, pyright all clean.
- `make test` — **1206 passed, 120 skipped, 4 xfailed, 1 xpassed**. The two `tests/evaluation/generate_modelcard` tests that fail when pyarrow is merely uninstalled now pass, and the faceting/visualization tests that assert on compiled-Vega output are unaffected (confirming vl-convert compiles the specs correctly). A parametrized regression test asserts the four horizon/time-period sum and mean plots carry no in-spec aggregate transform in their compiled Vega.
- **End-to-end with a real model.** Ran `chap eval` against [`auto_regressive_monthly_v2`](https://github.com/chap-models/auto_regressive_monthly_v2) on both `example_data/laos_subset.csv` and `example_data/hydromet_5_filtered.csv`; both produced valid evaluation NetCDFs. From the real eval output, rendered every `chap plot-backtest` type (`evaluation_plot`, `horizon_location_grid`, `predicted_vs_actual`, `sample_bias_by_horizon`, `sample_bias_by_time_period`) to **PNG, HTML, and SVG**, plus a full `chap generate-modelcard` (markdown + 10 plots) — all succeed and render correct charts (observed lines + forecast confidence bands). Confirmed at runtime that neither `vegafusion` nor `pyarrow` is imported.
- As expected from the trade-off, real-data specs are larger without pre-aggregation (e.g. `evaluation_plot.html` ~84 KB and `horizon_location_grid.png` ~1 MB on the Laos eval).

[CLIM-648]: https://dhis2.atlassian.net/browse/CLIM-648?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


knutdrand:
Not part of the diff, so leaving it here rather than inline: the three `alt.data_transformers.enable("default")` fixtures under `tests/evaluation/` are now obsolete and mildly harmful.

- `tests/evaluation/conftest.py:17` — `default_transformer` (module-scoped)
- `tests/evaluation/test_dataset_plot.py:19` — `default_transformer`
- `tests/evaluation/test_evaluation_plot.py:24` — `default_altair_transformer` (`autouse=True`)

They existed only to undo the module-level `alt.data_transformers.enable("vegafusion")` in `chap_core/plotting/{yaml_plot,backtest_plot,dataset_plot}.py`. This PR replaces those with `alt.data_transformers.disable_max_rows()`, so there is nothing left to opt out of.

Worse, they now actively fight that: `enable("default")` resets the transformer options to `{}`, which re-imposes `max_rows=5000` process-wide, and none of the fixtures restore the previous state on teardown. The `autouse` one in `test_evaluation_plot.py` means every test in that module leaves the process in that state. Concretely: after `test_evaluation_plot.py` runs, any later test in the same session that renders a chart with >5000 rows raises `MaxRowsError` even though production would not, since production explicitly disables the cap at import.

Nothing hits this today (collection order is deterministic and `pytest-randomly` isn't installed), so it's latent rather than a live failure. Suggest just deleting all three fixtures and their parameter usages as part of this cleanup.

mortenoh:
Good catch, and confirmed on both counts. Fixed in 9168ec68.

Verified the state leak locally on altair 6.2.1:

```
alt.data_transformers.disable_max_rows()  -> options {'max_rows': None}
alt.data_transformers.enable('default')   -> options {}
```

So `enable("default")` does reset options rather than merge, re-imposing the 5000-row cap process-wide with no teardown to restore it.

Deleted all three fixtures and their parameter usages:
- `tests/evaluation/conftest.py` — `default_transformer` (plus the now-unused `import altair`)
- `tests/evaluation/test_dataset_plot.py` — local `default_transformer` (plus its unused `import altair as alt`)
- `tests/evaluation/test_evaluation_plot.py` — `default_altair_transformer` (the `autouse` one; `alt` import stays, it is still used for the `alt.Chart` assertion)

Usages removed from `test_backtest_plot.py` (10) and `test_causal_plot.py` (5) as well.

`make lint` clean; `make test` 1209 passed, 117 skipped, 4 xfailed, 1 xpassed.

knutdrand (approved):

PR #470, opened 2026-06-26
chore(architecture): C4 architecture model with multi-renderer diagram export (CLIM-791)

mortenoh:
## What

Adds a code-as-truth **C4 architecture model** of the CHAP platform under `architecture/`, aimed at devops and power users (how the system runs/deploys, not source internals). It also adds a small set of companion **contributor documentation pages** in the mkdocs site that explain the same system in prose + Mermaid.

Authoring is in **Structurizr DSL** (`architecture/workspace.dsl`, the single source of truth). The model renders in an interactive viewer and exports to committed PNGs, plus optional trials of other renderers.

## C4 in one minute

[C4](https://c4model.com) (by Simon Brown) describes software as a hierarchy of **zoom levels**, so each diagram answers one question at one altitude instead of one diagram trying to show everything:

1. **Context** — the system as a single box among its users and neighbouring systems. "Who uses it and what does it talk to?"
2. **Containers** — the separately deployable/runnable parts inside a system (apps, services, datastores). "What are the moving pieces and how do they communicate?" (A "container" here means a runtime unit — a process/service/DB — not a Docker container.)
3. **Components** — the major building blocks inside one container. "How is this service organised internally?"
4. **Code** — class-level detail; usually skipped and left to the IDE.

Same elements, increasing zoom — you drill from a box on one level into its own diagram on the next. C4 is notation-agnostic (icons/colours are optional), and **dynamic views** add numbered, sequence-style flows over the same elements.

This model maps onto that directly:

| C4 level | Views here |
| --- | --- |
| Context (landscape) | `L1_Landscape` |
| Containers | `L2_ChapCore`, `L2_Chapkit` |
| Components | `L3_CoreAPI`, `L3_CoreWorker`, `L3_ChapkitService` |
| Dynamic (flows) | `Flow_IngestDataset`, `Flow_Backtest`, `Flow_Prediction` |

## Levels (views)

- **L1 — System landscape**: DHIS2 <=> CHAP Core => model services, every actor.
- **L2 — Containers**: CHAP Core (REST API, Celery worker, Redis/Valkey, PostgreSQL, CLI) and a chapkit model service (Service API, SQLite store, web console).
- **L3 — Components**: CHAP Core REST API, CHAP Core worker, and the chapkit Service API internals.
- **Flows**: ingest dataset, run a backtest, and run a prediction with review + optional DHIS2 write-back.

## Companion documentation (mkdocs)

The raw C4 model stays standalone under `architecture/`, but a few prose + Mermaid pages that explain the same system now live in the mkdocs site, under **Contributors**:

- **CLI vs REST API** (`docs/contributor/cli_vs_rest_api.md`) — makes the two entry points' shared vs unique capabilities explicit: a Mermaid diagram (one shared `chap_core` core, two entry points) plus a capability matrix (`CLI` / `REST API` / shared module) so a reader can pick the right entry point for a task.
- **Modeling internals: in-tree vs external ML** (`docs/contributor/modeling_internals.md`) — draws the boundary between the orchestration/evaluation code inside `chap_core` and the external forecasting models. Classifies in-tree code as genuinely-ML / ML-tooling / not-ML, and lists the ML libraries actually used (and the deep-learning ones deliberately absent).
- **Architecture model (C4)** (`docs/contributor/architecture_model.md`) — the full C4 model (L1/L2/L3 + the three flows) rendered as native, theme-aware Mermaid directly in the docs site. It is **generated from `architecture/workspace.dsl`** (via `make architecture-export-docs` + `architecture/mermaid_to_docs.py`, which collapses the Structurizr Mermaid export's HTML labels to plain text), so it cannot drift from the canonical model.
- All are wired into the Contributors nav and cross-linked from the existing `architecture_diagrams.md` guide.

## Direction-of-flow facts the model encodes (verified against the code)

- Predictions are **collected/stored in CHAP** and only written to DHIS2 by the Modelling App **after human review** — CHAP never writes to DHIS2 itself.
- Models are published as either an **MLproject repo** (run in-process by the worker) or a **chapkit service** (called over HTTP). chapkit is modelled in depth (routers, ML manager, async scheduler, model runner, SQLite store, web console, registration handshake).
- chapkit services are **[0..\*]** — one per model, zero-to-many — the now-preferred path over MLproject.

## Tooling (Makefile)

- `make architecture` — interactive Structurizr viewer (zoom/pan/drill-down) at **http://localhost:6080**.
- `make architecture-validate` — validate the DSL.
- `make architecture-export` — render every view to `architecture/diagrams/*.png` (also pre-warms viewer thumbnails). Exports are reproducible (no timestamp footer).
- `make architecture-export-mermaid` / `-plantuml` / `-export-likec4` — render the same model with other tools into `architecture/diagrams/<renderer>/` for side-by-side comparison.
- `make architecture-likec4` — interactive LikeC4 viewer at **http://localhost:6081**.
- `make architecture-export-docs` — regenerate the native-Mermaid C4 model page (`docs/contributor/architecture_model.md`) from `workspace.dsl`.
- `make docs-serve` — live-reloading mkdocs preview at **http://localhost:8000**.

Viewer ports were moved off `8080`/`5180` to `6080`/`6081` (browser-safe — 6000 is a browser-blocked X11 port — and not macOS AirPlay-reserved); both interactive viewers echo their URL on start. Mermaid diagrams in the docs get pan/zoom + fullscreen via the `mkdocs-panzoom-plugin` (added as a dev dependency).

## Notes

- The Structurizr/LikeC4 **model** (raw DSL + PNG exports) is kept standalone under `architecture/` and is not part of the mkdocs build; only the companion prose + Mermaid pages ship in the docs site.
- Structurizr's cloud-hosted `theme default` (EOL 30 Sep 2026) was removed in favour of explicit styles; the `local`/`validate`/`export` tooling we use stays free and open source.
- Runtime artifacts (`workspace.json`, `.structurizr/`, `exports/`, `_site/`, `node_modules/`) are gitignored; the model, scripts, README, and rendered PNGs are tracked.

## Model-vs-code verification (latest)

Ran a multi-agent review verifying every claim in **both** models against the real `~/dev/chap-sdk` repos (chap-core, chapkit, chap-frontend, and the model repos), with each discrepancy adversarially re-checked against source. Resulting corrections:

- **Orchestrator** is a Redis-only TTL registry — the model-template DB sync is performed by **v1** (reading the live set via the orchestrator), so the inaccurate `orchestrator -> PostgreSQL` edge was removed.
- Modelled the **v2 read-only reverse proxy** to live chapkit services (artifact/config/job browsing).
- chapkit registration/health serves `/health` and `/api/v1/info` (the service identity chap-core reads to decide dispatch; not `/system`, and not `/api/v1/system`, which is a separate system-stats endpoint); Redis is also noted as the Celery **result backend**.
- **LikeC4 port** brought back in sync with the canonical DSL: external systems styled grey, the three dynamic **Flow** views ported (LikeC4 now exports 9 PNGs, matching the other renderers), and relationship technology labels restored.
- LikeC4 viewer/export now serve in single-page mode (`serve -s`) so deep-linked views return `200` instead of `404`.

Review follow-ups (second pass):

- **Worker resolves the live chapkit URL from Redis directly**, by constructing its own `Orchestrator` over Redis in-process - it does not call the REST API's orchestrator component. The edge is now `worker -> Redis`, so the diagram no longer implies a worker-to-API runtime dependency.
- The renderer toolchain (Structurizr image, LikeC4, `serve`, mermaid-cli, PlantUML) is **version-pinned** in the `Makefile` so the committed PNGs stay reproducible across re-exports.
- `export-likec4.js` now waits for a rendered diagram node before screenshotting, so a bad view id / route change fails loudly instead of committing an empty page.

Review follow-ups (third pass):

- README "Under the hood" examples are pinned to `structurizr/structurizr:2026.05.22`, matching the `Makefile` so the docs do not undercut the reproducibility guidance.

Review follow-ups (fourth pass):

- The ingest and backtest **dynamic flows** now show the REST API only validating and queueing, with the **worker** persisting: ingest saves DataSet + Observations in the worker; backtest queues with the existing `model_id` (no ConfiguredModel is created), and the worker saves the dataset, loads the configured model by id, runs folds and writes results.

Draft for review of the model and the diagrams.


architecture/diagrams/plantuml/L1_Landscape.png
jenniferarnesen, line 1: Should chapkit model services and Model source repos show some kind of relationship? chapkit is in a way an interface to the model repo, right?

Is there a way to arrange the diagram to reduce the criss-crossed lines?

Chap core <--> chapkit model services shows 2-way communication, but 2-way comms isn't shown between other boxes.

mortenoh: chapkit has no dependency on a model repo, so that would be wrong i feel. I _could_ be living in a repo somewhere, but that has nothing to do with chapkit itself (unlike a worker, which is heavily dependent on repos for running models).

I'll try to arrange it a bit, but its mostly automatic layout (can be tweaked a bit, but not much)

only chap-core and chapkit has 2 way communication, no other model types (chapkit do self-registration on chap-core, thats why there is a bi-directional communication)

knutdrand:
Two comments on the docs:

**1. The flows on the main page should be `create-backtest-with-data`, not dataset creation + backtest creation as two steps.**

The Modelling App does not create a dataset first and then create a backtest from it — it posts observations and geojson in a single call to `POST /v1/analytics/create-backtest-with-data`. Presenting a separate `make-dataset` ingest journey alongside the backtest flow (the `Flow_IngestDataset` view in `architecture/workspace.dsl`, rendered as "Flow - Ingest dataset" in `contributor/architecture_model.md`) makes the two-step path look like the normal frontend flow, which it isn't.

Same for predictions: the frontend uses the inline-data endpoint `POST /v1/analytics/make-prediction`, so the prediction flow's first step ("Request prediction") should name that endpoint rather than implying a dataset is created up front.

**2. `contributor/modeling_internals.md` should be clearer that chap-core does not do the modelling.**

Models live independently outside chap-core; chap-core only provides the tooling around them (running, evaluating, storing and serving results). There are parts of chap-core that technically do modelling, but that should not be the main narrative — a reader should come away understanding that chap-core is not where models are implemented.

ivargr:
Had some issues running `make architecture`, which is probably a linux thing, resulting in:

```
   ERROR com.structurizr.configuration.Configuration -- Data directory
      /usr/local/structurizr is not writable
```

Claude's explanation:

  `1b0eb5bb` added `$(DOCKER_AS_HOST_USER)` to the export targets, but the viewer
  target (Makefile:131) and the Structurizr container inside `architecture-export`
  (Makefile:141) still run as the image's own non-root uid, so they cannot write
  `workspace.json` / `.structurizr/` into the bind mount. Docker Desktop on macOS
  virtualises this away, which is probably why it slipped through. Adding
  `$(DOCKER_AS_HOST_USER)` to both fixes it (verified: viewer returns 200 on
  `/workspace/1` and `workspace.json` is written as the host user). The README's
  "Under the hood" examples need the same flag so they stay honest.

ivargr:
Checked this out locally on Linux — `make architecture` fails before the viewer starts:

```
ERROR com.structurizr.configuration.Configuration -- Data directory /usr/local/structurizr is not writable
```

`1b0eb5bb` added `$(DOCKER_AS_HOST_USER)` to the export containers, but the viewer target (`Makefile:131`) and the Structurizr instance inside `architecture-export` (`Makefile:141`) still ran as the image's own non-root uid, so they could not write `workspace.json` / `.structurizr/` into the bind mount. Docker Desktop on macOS virtualises this away, which is probably why it slipped through.

Pushed a fix (b81c09f4): `--user` on all three Structurizr containers, plus the same flag in the README's "Under the hood" examples so they match the make targets. Verified the viewer now returns 200 on `/workspace/1` and writes `workspace.json` as the host user; `make architecture-check-docs` and `make docs` (strict) both pass.

Three other things worth addressing before review:

1. **The PR description is out of date.** It still documents the committed PNGs, the LikeC4 model, `make architecture-likec4` / `architecture-export-likec4` and port 6081, all removed in `8c536d82`. A reviewer following it will look for targets that no longer exist.

2. **panzoom does not reach the published site.** `chap-site` imports these docs via `mkdocs-multirepo` (`!import .../chap-core`) and uses its own plugin list. It has superfences/mermaid, so the diagrams render, but not `mkdocs-panzoom-plugin` — so pan/zoom and fullscreen are local-only. That is exactly what `L1_Landscape` and `L3_ChapkitService` need at their size. The plugin has to be added to `chap-site`'s `requirements.txt` and `mkdocs.yml` too.

3. **Docs CI now depends on Docker Hub.** `make architecture-check-docs` pulls a 431 MB Structurizr image on every docs PR and runs before `make docs`, so a Docker Hub rate limit blocks the docs build and deploy for PRs that never touch the model. Could that step be gated on `paths: architecture/**`?

The model itself checks out against the code — I spot-verified the v2 proxy being GET/HEAD-only (`rest_api/v2/routers/proxy.py:49`), Redis as both Celery broker and result backend (`rest_api/celery_tasks.py:76`), `/api/v1/info` as the chapkit identity endpoint (`models/chapkit_rest_api_wrapper.py:92`), and the v1 model-template sync going through the orchestrator (`rest_api/v1/routers/crud.py:88`). The CLI commands in the capability matrix all exist.

Minor: the node ids in the generated Mermaid page are sequential numbers, so inserting an element mid-`workspace.dsl` renumbers everything below it and churns the whole page in the diff.

ivargr (approved):

PR #509, opened 2026-08-11
ci: run the test matrix on the platforms it declares

mortenoh:
Stacked on #508 because it touches the same file. GitHub retargets this to `master` automatically when #508 merges.

Draft on purpose: the point is to find out what Windows and Intel macOS actually do.

## The bug

`ci-test-python-install.yml` declares `matrix.os: [ubuntu-latest, windows-latest, macos-latest]` but hardcodes `runs-on: ubuntu-latest`, and has since the initial commit (`71147aff`). All three legs run on ubuntu-24.04 — confirmed from the `test (macos-latest, 3.13)` job log of run [31413568791](https://github.com/dhis2-chap/chap-core/actions/runs/31413568791), which reports `Image: ubuntu-24.04`.

So two required status checks currently assert cross-platform coverage that does not exist, and ~7 minutes of runner time per PR goes to duplicate ubuntu runs.

The intent was clearly there once: `tests/conftest.py:30` still defines `IS_WINDOWS = os.name == "nt"` with the comment "Don't use pytest-celery if on windows", and its only use is commented out at line 72.

## The change

`runs-on: ${{ matrix.os }}`, plus a fourth leg for Intel macOS:

| Label | Platform | Check name |
|---|---|---|
| `ubuntu-latest` | Linux x64 | `test (ubuntu-latest, 3.13)` (required, unchanged) |
| `windows-latest` | Windows x64 | `test (windows-latest, 3.13)` (required, unchanged) |
| `macos-latest` | macOS arm64 (macOS 26) | `test (macos-latest, 3.13)` (required, unchanged) |
| `macos-15-intel` | macOS x86_64 | `test (macos-15-intel, 3.13)` (new) |

macOS amd64 vs arm64 is a recurring source of issues, and `macos-latest` is Apple Silicon, so Intel needs its own label. Both are standard runners and [free and unlimited on public repositories](https://docs.github.com/en/actions/reference/runners/github-hosted-runners).

`fail-fast: false` so one platform failing does not cancel the others — otherwise the first failure hides the rest.

## Notes

- The three existing job names are byte-identical, so the branch ruleset needs no edit. If the new Intel leg proves useful it can be added as a required check afterwards.
- Docker and redis tests already self-skip when those are unavailable (`tests/test_cli.py:12`, `tests/runners/test_runners.py:30`, `tests/external/test_docker.py`, `tests/test_alembic_migrations.py:66`, and the `redis_available` fixture), which covers most of what the non-Linux runners lack.
- Wall clock should be unaffected: these legs run in parallel with `test-all`, which takes 13 minutes. Anything under that is free.
- `pytest_plugins = ("celery.contrib.pytest",)` in `tests/conftest.py` is unconditional. That is the most likely Windows breakage.


mortenoh:
## Results from the first real run

Run [31474303776](https://github.com/dhis2-chap/chap-core/actions/runs/31474303776). `fail-fast: false` earned its place immediately — two legs failed for completely unrelated reasons.

| Leg | Result | Detail |
|---|---|---|
| `ubuntu-latest` | pass | |
| `macos-latest` (arm64) | pass | |
| `macos-15-intel` (x64) | **fail** | `uv sync` cannot install |
| `windows-latest` | **fail** | installs in 21s, 62 of 1331 tests fail in 20m55s |

### Intel macOS is not installable, and cannot be made so

`llvmlite` publishes **arm64-only macOS wheels** from 0.46.0 onward. There is no `macosx_*_x86_64` wheel, only an sdist, so uv falls back to compiling it and CMake fails with `Could not find a package configuration file provided by "LLVM"`.

| llvmlite | cp313 macOS x86_64 | cp313 macOS arm64 |
|---|---|---|
| 0.45.1 | yes | yes |
| 0.46.0 - 0.49.0rc1 | **no** | yes |

The chain is `chap-core` -> `pyts` + `stumpy` -> `numba` -> `llvmlite`, and `stumpy`/`pyts` are in `[project] dependencies` (pyproject.toml:63-64), not optional extras. So this is not a dev-environment quirk: **chap-core cannot be installed on an Intel Mac at all**, by anyone, on Python 3.13.

Rosetta does not help — it translates x86_64 to arm64 and only exists on Apple Silicon; it cannot make an Intel Mac run arm64 wheels. Docker does work for local development (linux/amd64 gets manylinux wheels) but cannot fix this CI leg: GitHub's macOS runners have no Docker, and running Linux containers on a macOS runner would just re-test what `ubuntu-latest` covers.

Proposed: drop `macos-15-intel` and document that Intel Macs must use the Docker path. Keeping a permanently red required-adjacent leg has no value when the cause is upstream and deliberate.

### Windows is worth keeping: 1269 pass, 62 fail

| Count | Cause |
|---|---|
| 25 | `tests/test_documentation.py:48` — `SKIP_FILES` compares `str(f).startswith(...)` against forward-slash paths, but `str(Path)` on Windows yields `docs\contributor\...`. **The entire skip list silently does nothing on Windows**, so ~20 files explicitly marked untestable get executed. |
| 19 | `tests/test_ci_change_scope.py` shelling out to `bash` — my own regression, fixed in e8efb265 |
| 7 | `tests/integration/rest_api/test_services_router.py` — 404s and a lifecycle assertion |
| 4 | genuine docs code-block failures |
| ~7 | misc: `PermissionError` on tempfiles in `test_polygons.py`, `test_get_temp_dir.py` asserting a literal `/tmp/chap/temp`, docker image 404s |

The `SKIP_FILES` one is a one-line fix (`f.as_posix()`) that clears 25 of 62 on its own.

### Wall clock caveat

Windows took **20m55s**, which exceeds `test-all`'s ~13m. So unlike the other legs this one does extend PR wall clock, by roughly 8 minutes. Worth considering `-n auto` tuning or trimming what runs on Windows once it is green.

PR #512, opened 2026-08-11
feat: add opt-in API token to the chap-core API

mortenoh:
Adds an opt-in API token to the chap-core API, so a Chap Core instance reachable from outside an internal network is at least defensible.

Today everything under `/v1/**` and `/v2/**` is unauthenticated, including `DELETE /v1/crud/datasets/{id}`. That is deliberate, and authorization is enforced by DHIS2 in front via the Route API. But users have started asking to connect a cloud DHIS2 instance to a local Chap Core over a tunnel, which puts the API directly on the public internet.

Setting `CHAP_API_TOKEN` requires `Authorization: Bearer <token>` on every endpoint except `/health`, `/health/ready` and `/system/info`. Leaving it unset disables authentication entirely, so no existing deployment changes.

## Design notes

Enforcement is an ASGI middleware rather than a route dependency, so `/docs`, `/redoc` and `/openapi.json` are covered too. It is raw ASGI rather than `BaseHTTPMiddleware` so the streaming service proxy passes through untouched, and it is registered before `CORSMiddleware` so 401 responses keep their CORS headers and preflight is answered before the token check. A non-enforcing `HTTPBearer(auto_error=False)` app dependency declares the scheme in the OpenAPI spec so generated clients pick it up.

`/system/info` stays public and now reports `auth_required`, so the Modeling App can tell "server needs a token" apart from "server unreachable" before it has one.

## Chapkit self-registration

servicekit's `registration.py` can only send `X-Service-Key` and has no way to send `Authorization`, so enabling the token would otherwise 401 every self-registering model service and make it vanish from the model list. The gate therefore also accepts the token in `X-Service-Key`, and on `/v2/services` paths a configured `SERVICEKIT_REGISTRATION_KEY` is accepted there as well. That acceptance is confined to those paths, so a registration key cannot be used as a general-purpose credential against `/v1/crud/**`.

Consequence: with both secrets set, a chapkit service sending only its registration key registers successfully.

## Included beyond the token itself

- The v2 proxy no longer forwards `Authorization` upstream, which would have leaked the caller's token to every registered service.
- The two secret checks now share helpers with a timing-safe comparison; `verify_service_key` previously used `!=`.
- **Behaviour change:** a missing `X-Service-Key` now returns 401 instead of 422, consistent with an invalid key. Documented in `service-registration.md`.
- Startup logs a warning if the token is shorter than 32 characters. There is no rate limiting, so a short token is brute-forceable.

## Not included

Helm chart plumbing, and the Modeling App changes in `chap-frontend` (token field in settings, passing it via the DHIS2 Route, and reading `auth_required` to drive `ServerSecurityNotices`, which currently hardcodes "your server is not secure"). Both want linked issues.

## Verification

`make lint`, `make test` (1232 passed), `make test-docs` (118 passed) and `mkdocs build --strict` all pass. The gate was also checked against a real uvicorn server rather than only `TestClient`: `/health` 200, `/system/info` reports `auth_required: true`, `/openapi.json` and `/v2/services` 401 with `WWW-Authenticate: Bearer`, and both a valid bearer token and the token in `X-Service-Key` pass the gate.

Refs CLIM-854

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -5,3 +5,9 @@ POSTGRES_DB=chap_core
 
 # Valkey/Redis
 REDIS_PASSWORD=your_secure_redis_password
+
+# API authentication (optional). Unset means no authentication.
+# Generate with `openssl rand -hex 32`; use at least 32 characters.
+# CHAP_API_TOKEN=
+# Shared secret for chapkit service registration (optional, separate from CHAP_API_TOKEN).
+# SERVICEKIT_REGISTRATION_KEY=
--- a/chap_core/rest_api/app.py
+++ b/chap_core/rest_api/app.py
@@ -2,12 +2,14 @@ import logging
 import os
 import traceback
 
-from fastapi import FastAPI, Request
+from fastapi import Depends, FastAPI, Request
 from fastapi.middleware.cors import CORSMiddleware
 from fastapi.middleware.gzip import GZipMiddleware
 from fastapi.responses import JSONResponse
+from fastapi.security import HTTPBearer
 
 from chap_core import __version__ as chap_core_version
+from chap_core.rest_api.auth import ApiTokenMiddleware, warn_on_weak_token
 from chap_core.rest_api.common_routes import router as common_router
 from chap_core.rest_api.v1.rest_api import router as v1_router
 from chap_core.rest_api.v2.rest_api import router as v2_router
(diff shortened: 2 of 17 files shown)


ivargr (approved):

ivargr:
Just a small note: should x-service-key also be dropped so that it is not forwarded (like authorization)?

mortenoh:
Good catch, and yes it should. The API token gate accepts the token in `X-Service-Key` on any path, so a caller authenticating that way had it forwarded to the registered service by the proxy. Fixed in #513.

PR #527, opened 2026-08-19
fix: preserve model template and configuration provenance (CLIM-886)

edvinstava:
Fixes CLIM-886 (https://dhis2.atlassian.net/browse/CLIM-886).

## Summary

The seeding code had two bugs:

1. A new version in config/configured_models/*.yaml replaced the stored template row. Every backtest pointing to that row then showed the wrong model code.

2. Changes to a configuration in the same file had no effect, and CHAP gave no warning.

This PR makes model templates and configurations immutable. A change now creates a new row instead of replacing an existing one.

## Unique (name, version) instead of unique name

Previously, the name column was unique. A new version label therefore had to update the stored row, causing the first bug.

This PR changes the uniqueness constraints:

- ModelTemplateDB is unique on (name, version), and version cannot be empty.
- ConfiguredModelDB is unique on (model_template_id, name, configuration_digest).
- configuration_digest is a hash of the option values and additional covariates.

An is_live flag identifies the row that CHAP currently serves for a name. GET /v1/crud/model-templates returns only live rows, so existing clients continue to see one template per name.

Old rows retain their IDs. Backtests, predictions, and prediction setups reference rows by ID, so they continue to show the model code that was originally used.

## No update path

add_or_update_model_template no longer has an update parameter and never writes to an existing row. Similarly, add_configured_model does not write to an existing row.

The resulting behavior is:

- A new version label creates a template row. The new row becomes live, while the old row remains in the database.

- A changed configuration creates a configuration row. The new row becomes live.
- If the content changes without a version-label change, CHAP keeps the stored row and logs a warning naming the fields it ignored.

- Seeding the same template or configuration again reuses the stored row.
- A prediction setup retains its link to its configuration row by ID.

POST /prediction-setups/{id}/run now sends the configuration row ID to the worker. Previously, the worker looked up the configuration by name and could therefore run a different version.

## Models without a repository

For a Git template, CHAP resolves the ref—for example, @main—to a commit SHA during seeding. It stores the SHA in source_digest and reads the model metadata from the same commit.

A branch can move, but the stored row cannot change. If CHAP cannot resolve the ref, it skips that model and logs an error. The remaining models are still processed, and startup continues.

Chapkit services and the built-in naive model do not have repositories, so they do not have a source_digest. Their identity consists of the name and the version label reported by the service in info.version.

A Chapkit model developer must therefore provide a new info.version after changing the configuration schema. If the version remains unchanged, CHAP keeps the stored row and logs a warning. Previously, CHAP reread the configuration schema during every sync.

Rows seeded before this change also have no source_digest. Because a version is write-once, CHAP does not add the digest later. An empty digest correctly indicates that the revision is unknown.

## Migration

alembic/versions/b8c9d0e1f2a3_version_templates_and_configurations.py:

- Adds source_digest and is_live to ModelTemplateDB.
- Makes version non-nullable.
- Adds configuration_digest and is_live to ConfiguredModelDB.
- Calculates a digest for every stored configuration row.
- Replaces the two uniqueness constraints on name.

config/configured_models/README.md documents the new rules for operators.

## Databases first created by create_all

Some deployments were never initialized through Alembic: the generic startup migration and SQLModel.metadata.create_all built their schema, and the alembic_version table is empty. On those databases the migration chain previously failed early on duplicate columns, the error was swallowed, and the legacy UNIQUE(name) constraints survived. Seeding a second version of a template then hit an IntegrityError that the seeding loop also swallowed, so the model silently disappeared.

Startup now detects this state (no Alembic revision, but modeltemplatedb exists) and stamps the database at the revision preceding the versioning migration. The versioning migration then runs and replaces the legacy unique-name constraints. Its constraint creation checks for existing constraints first, so it is also idempotent on fresh databases where create_all already produced the new schema.

The stamp assumes the generic migration plus create_all provide the schema of the stamped revision. Cleanups from the skipped revisions (for example, dropping the old configuredmodelwithdatasource table) are not applied to such databases; they are stale leftovers and do not affect behavior.

## Seeding failures are no longer silent for database errors

The per-model exception handlers in seed_configured_models_from_config_dir now re-raise SQLAlchemyError after rolling back, so a schema problem fails startup loudly instead of silently omitting a model. Non-database errors (for example, an unreachable repository or Chapkit service) still skip that model and let startup continue.


chap_core/database/model_template_seed.py
```diff
@@ -14,18 +15,26 @@
...
 
 
 def add_model_template_from_url(
     url: str, session_wrapper: SessionWrapper, version: str, name_override: str | None = None
 ) -> int:
-    model_template_config = ExternalModelTemplate.fetch_config_from_github_url(url)
+    # A git template must point to one revision, so an unknown ref is an error.
+    source_digest = resolve_commit_sha(url)
+    if source_digest is None:
+        raise ValueError(f"Could not resolve an immutable source digest for model template URL {url!r}")
```
knutdrand, line 33: **High - this `ValueError` aborts startup, it is not skipped.**

The git branch of `seed_configured_models_from_config_dir` (line 128 in this file) has no `try/except` around this call - only the chapkit branch does, and it catches `TimeoutError` only. `create_db_and_tables` does not guard the seeding call either (`database.py:573`), and it runs from gunicorn's `on_starting` hook (`gunicorn.conf.py:63`).

Scenario: an operator adds a yaml entry whose last version is `@main` (the documented `nightly_build` pattern in `default.yaml`). GitHub answers 403 - the unauthenticated API limit is 60/hr/IP, easy to hit in a container restart loop - or the branch was renamed, so `resolve_commit_sha` returns `None`. This raises, every remaining model *and* the naive model template are never seeded, the final `session.commit()` is skipped, and the API never boots.

The PR description says CHAP "skips that model and logs an error. The remaining models are still processed, and startup continues" - the code does not do that. Catching `ValueError` alongside `TimeoutError` in the seeding loop (and logging + `continue`) would match the description.

chap_core/database/database.py
```diff
@@ -113,24 +110,59 @@ def _add_model_template(self, model_template: ModelTemplateDB) -> int:
...
-        # convert yaml config to model template db object and add to db
+    def add_model_template_from_yaml_config(
+        self, model_template_config: ModelTemplateConfigV2, source_digest: str | None = None
+    ) -> int:
         d = model_template_config.model_dump()
         info = d.pop("meta_data")
         d = d | info
+        # The version is part of the identity, so it cannot be empty.
+        if not d["version"]:
+            raise ValueError(f"Model template {d['name']!r} must declare a version")
```
knutdrand, line 193: **Medium - same startup-abort class as the git path.**

The chapkit branch of `seed_configured_models_from_config_dir` calls this at `model_template_seed.py:106`, inside a `try` that catches `TimeoutError` only (`model_template_seed.py:116`). A chapkit service that reports an empty or null `info.version` now raises here, aborting the whole seeding loop and gunicorn startup, where before it seeded fine.

knutdrand (commented):
## Code review

Reviewed hunk by hunk against `master`. `make lint` passes; `tests/test_database.py` and `tests/test_github.py` pass locally (32 passed, 2 skipped). The alembic PG-container test was not run (no Docker available).

Eight findings, seven of them left as inline comments. The remaining one is on a line outside the diff:

**`chap_core/rest_api/v1/routers/crud.py:762` (medium) - `list_model_templates` has no `is_live` filter.** The PR description says "GET /v1/crud/model-templates returns only live rows, so existing clients continue to see one template per name", but the endpoint still does a plain `select(ModelTemplateDB)`. The new test `test_re_registered_service_with_new_version_adds_new_template` asserts `len(matching) == 2`, so after the first version bump any client that keys the template list by `name` (a model picker, for instance) renders duplicate entries. Either the filter or the PR description needs to change.

The two startup-abort findings (inline, on `model_template_seed.py:28` and `database.py:162`) look worth fixing before merge - both can take the API down at boot.

chap_core/rest_api/v1/routers/crud.py
```diff
@@ -119,17 +119,16 @@ def _sync_live_chapkit_services(session: Session, orchestrator=None) -> set[str]
...
+                # metadata if its config schema is temporarily unavailable. A later
+                # sync can then create the template with its full user options.
                 try:
                     client = CHAPKitRestAPIWrapper(service.url, timeout=5)
                     schema = client.get_config_schema()
                     user_options = _parse_user_options_from_config_schema(schema)
                 except Exception:
-                    logger.debug("Could not fetch config schema from %s", service.url)
+                    logger.debug("Could not fetch config schema from %s, will retry next sync", service.url)
+                    continue
```
knutdrand, line 135: **Medium - this silently drops services that were previously registerable.**

Before this change the service was still registered with `user_options = {}`, which is an explicitly supported mode - see `add_configured_model`'s own comment that chap-core stores `user_option_values={}` as a "use chapkit defaults" sentinel.

Scenario: a chapkit build that does not expose `/api/v1/configs/$schema`, or one behind auth. It is now never registered at all, so the model never appears in `/model-templates` or `/configured-models`, and the only trace is a `logger.debug` line an operator will not have enabled. At minimum this wants `logger.warning` with the exception attached.

chap_core/database/database.py
```diff
@@ -113,24 +110,59 @@ def _add_model_template(self, model_template: ModelTemplateDB) -> int:
...
+            template_id = self._return_model_template_id(model_name, existing_template)
+            # Show the template again, but do not change its contents.
+            if existing_template.archived:
+                existing_template.archived = False
+                self.session.commit()
         else:
-            return self._add_model_template(model_template)
+            # add_model_template_from_url gives git templates a digest. Other sources give None.
+            template_id = self._add_model_template(model_template)
+        self._make_live_template_version(model_name, template_id)
```
knutdrand, line 182: **Medium - the new version goes live before it has any configured model.**

`_select_live_configured_models` requires `ModelTemplateDB.is_live == True` (`database.py:245`), so the moment this marks the new template version live, the previous version's configured models stop being visible - and the new version has none yet.

Scenario: a chapkit service bumps `info.version`. The new template row is created and made live here; `_sync_chapkit_configured_models` then hits a transient failure in `client.list_configs()` and returns early (`crud.py:227`). The model now has zero live configured models: it disappears from `GET /v1/crud/configured-models`, and `get_configured_model_by_name` raises "not found" for every backtest or prediction created by name, until a later sync succeeds. The same window exists on the yaml path if `add_configured_model` raises, for example when `validate_user_options` rejects the new template's schema.

Making the version live only after its configured models are committed would close the window.

chap_core/external/github.py
```diff
@@ -26,6 +29,26 @@ def parse_github_url(github_url) -> GithubUrl:
...
+def resolve_commit_sha(github_url: str) -> str | None:
+    """Resolve the ref in a github url to the commit sha it points to.
+
+    A branch ref such as ``@main`` can move, but a sha cannot. A url that already
+    has a full sha needs no lookup. Returns None if the lookup fails.
+    """
+    parsed = parse_github_url(github_url)
+    if COMMIT_SHA_PATTERN.fullmatch(parsed.commit):
+        return parsed.commit
+    api_url = f"https://api.github.com/repos/{parsed.owner}/{parsed.repo_name}/commits/{parsed.commit}"
```
knutdrand, line 41: **Low/medium - unauthenticated, uncached GitHub call on every startup.**

This runs for every non-40-char-lowercase-hex ref at seeding time, including refs whose `(name, version)` row already exists - `add_or_update_model_template` then discards the result because a version is write-once. Unauthenticated `api.github.com` allows 60 requests/hr/IP.

`COMMIT_SHA_PATTERN` also does not match uppercase or abbreviated SHAs, so a config pinned as `@A1B2...` or a 7-char SHA triggers an avoidable call that returns a *different* string than what was pinned.

Combined with the `ValueError` on the caller side, a GitHub rate-limit response turns into a failed boot.

chap_core/rest_api/data_models.py
```diff
@@ -175,7 +175,11 @@ class ModelTemplateRead(DBModel, ModelTemplateInformation, ModelTemplateMetaData
         default=None, description="JSON-schema-like dict describing the template's user-configurable options."
     )
     required_covariates: list[str] = Field(default=[], description="Covariate names the template must be given to run.")
-    version: str | None = Field(default=None, description="Template version string, typically a git tag or commit sha.")
+    version: str = Field(description="Version label of this template.")
```
knutdrand: **Low - required `version` plus a silently-swallowed migration is a 500.**

`version` went from `str | None = None` to a required `str` here and in `model_spec_tables.py:42`, while `_run_alembic_migrations` swallows every exception (`database.py:540-543`, "Don't raise - allow system to continue if Alembic fails").

If the new migration fails on a deployment - for instance the pre-existing unique constraint is not actually named `modeltemplatedb_name_key`, so the backfill / NOT NULL step errors - `version` stays NULL on old rows and `GET /v1/crud/model-templates` and `/configured-models` raise a pydantic `ValidationError` -> 500, where before they returned `version: null`. The silent-migration-failure path is what makes this reachable in production.

chap_core/database/database.py
```diff
@@ -274,15 +336,14 @@ def get_configured_models(self) -> list[ModelSpecRead]:
         return configured_models_read
 
     def get_configured_model_by_name(self, configured_model_name: str) -> ConfiguredModelDB:
-        try:
-            configured_model = self.session.exec(
-                select(ConfiguredModelDB).where(ConfiguredModelDB.name == configured_model_name)
-            ).one()
-        except sqlalchemy.exc.NoResultFound:
+        configured_model = self.session.exec(
+            self._select_live_configured_models().where(ConfiguredModelDB.name == configured_model_name)
```
knutdrand, line 375: **Low - the error message below contradicts this lookup.**

The lookup is now restricted to live rows, but the "not found" message on line 343 still does `select(ConfiguredModelDB.name)` over *all* rows. After a version bump the message can read `Configured model with name X not found. Available names: [..., 'X', ...]`, listing the same name once per superseded row - actively misleading during exactly the failure described on `add_or_update_model_template`. Using `self._select_live_configured_models()` names there would fix it.

chap_core/database/database.py
```diff
@@ -155,42 +212,72 @@ def add_configured_model(
...
+                ConfiguredModelDB.configuration_digest == digest,
+            )
+        ).first()
         if existing_configured:
-            logger.info(f"Configured model with name {name} already exists. Returning existing id")
-            return cast("int", existing_configured.id)
+            logger.info(
+                f"Configured model {name} with an identical configuration already exists. Returning existing id"
+            )
+            configured_model_id = cast("int", existing_configured.id)
```
knutdrand: return configured_model_id here to avoid wrapping rest of function in else

chap_core/database/database.py
```diff
@@ -155,42 +212,72 @@ def add_configured_model(
...
-        # stores user_option_values={} as a "use chapkit defaults" sentinel.
-        # The local heuristic-based validator wrongly flags any default_factory
-        # field as required (no literal "default" key in the schema), so skip it.
-        if not uses_chapkit:
-            configured_model.validate_user_options(configured_model)
-        # configured_model.validate_user_options(model_template)
-        logger.info(f"Adding configured model: {configured_model}")
-        self.session.add(configured_model)
+        self._make_live_configured_model(model_template_id, name, configured_model_id)
+        self._make_live_template_version(template_name, model_template_id)
```
knutdrand, line 262: This makes live the old version if configured model exists. bug I think

chap_core/rest_api/v1/routers/crud.py
knutdrand: @mortenoh pulling you in on this one since you know the chapkit side.

Question: is `/api/v1/configs/$schema` guaranteed to be present on every chapkit build we expect to talk to? Two cases I am unsure about:

1. An older chapkit that predates the schema endpoint.
2. A model with no configurable options at all - does chapkit still serve a schema there, or 404?

If either can happen in practice, this `continue` is a regression: those services used to register with `user_options = {}` (the documented "use chapkit defaults" sentinel) and now get dropped entirely, so they never show up in `/model-templates` or `/configured-models` at all. Not "until the next sync" - the next sync takes the same branch.

If chapkit always serves the endpoint, then the only real case left is a transient failure and this is fine as-is, though it should still be `logger.warning` with the exception attached rather than `logger.debug`.

alembic/versions/b8c9d0e1f2a3_version_templates_and_configurations.py
```diff
@@ -0,0 +1,134 @@
...
+
+    # Version is part of the identity, so it cannot be null.
+    op.execute(
+        sa.text("UPDATE modeltemplatedb SET version = :version WHERE version IS NULL").bindparams(
+            version=LEGACY_UNVERSIONED_VERSION
+        )
+    )
+    op.alter_column("modeltemplatedb", "version", existing_type=sa.String(), nullable=False)
+    op.execute("ALTER TABLE modeltemplatedb DROP CONSTRAINT IF EXISTS modeltemplatedb_name_key")
+    op.create_unique_constraint("uq_modeltemplatedb_name_version", "modeltemplatedb", ["name", "version"])
```
knutdrand: **Medium - this constraint swap only ever happens on the Alembic path, and there are deployments that never take it.**

_Disclosure: this was found by Claude in an automated review pass, and I am passing it on because the reasoning looks sound to me. It is derived from reading the code paths - it has not been reproduced against a real legacy database, so please sanity-check the premise before acting on it._

The chain as I read it:

- `_run_generic_migration` adds missing columns but never touches constraints.
- `SQLModel.metadata.create_all` skips tables that already exist, so it will not add `uq_modeltemplatedb_name_version` to an existing `modeltemplatedb`.
- So the old `UNIQUE(name)` is only dropped if this migration actually runs.

For a database that was first created by `create_all` rather than by Alembic, the chain does not get that far: `a1b2c3d4e5f6` unconditionally re-adds the `archived` column that `create_all` has already created, and the resulting error is swallowed by the `except` in `_run_alembic_migrations` (`chap_core/database/database.py:565`). The old `UNIQUE(name)` survives.

After that, seeding any second version of a template name hits an IntegrityError - and the new blanket `except Exception: session.rollback(); continue` in `seed_configured_models_from_config_dir` swallows that too. Net effect on such a deployment: the model silently vanishes from the platform instead of failing loudly at boot.

The swallow-everything handler in the seeding loop is what turns this from a loud startup failure into a silent one, so it may be worth narrowing that handler regardless of what we conclude here.

chap_core/database/database.py
edvinstava: This was intentional, handling reverting to an older version. If the config is changed back to v1 after v2 was seeded, the configured model already exists, so without these calls the re-seed v2 stays live (instead of reverting to v1). 

But it does mean any call to add_configured_model makes it the live version, including POST /configured-models, so a user saving a config against an old template version would make that old version live for everyone. Should I move the _make_live_template_version call out of add_configured_model and into the seeding flow, so only seeding switches live versions? Or what do you think?

knutdrand: ah ok, I misread the existing_configured logic. Maybe just give two different names to the two interpretations of the existing_configured, then were good

knutdrand (approved):

PR #528, opened 2026-08-21
fix: make backtest failures readable and non-fatal per split

mortenoh:

Two independent robustness fixes surfaced by a reported EWARS backtest failure (9 provinces, 10 years, 5 covariates), where the underlying model crash was both unreadable and needlessly destructive.

The model bug itself -- a rank-deficient dlnm lag basis at `n_lags = 3` -- lives in the model repos and is not addressed here; it is tracked in [CLIM-929](https://dhis2.atlassian.net/browse/CLIM-929).

## 1. `run_command` reported output as concatenated byte values

`chap_core/runners/command_line_runner.py` called `process.communicate()` a second time after the first call had already consumed the streams, then embedded the result with `''.join(map(str, streamdata))`. Iterating `bytes` yields `int`, so each byte was rendered as a decimal number:

```
>>> out, err = p.communicate()
>>> ''.join(map(str, p.communicate()[0]))
'10410510'          # this is b'hi\n'
```

In the reported failure this produced a 12,167-digit run of concatenated ASCII codes carrying no information that the decoded output did not already contain. Combined with the exception being re-wrapped into `ModelFailedException` and printed in both halves of the `__cause__` chain, the R log appeared four times in what should have been a ~50 line report.

The redundant `communicate()` call is dropped and the message is built once from the already decoded stdout/stderr.

## 2. One failing split discarded the whole backtest

`chap_core/assessment/prediction_evaluator.py:backtest` let a per-split `predict` exception escape the generator. A model that fails on a single split -- numerical instability, a degenerate window -- took all the other splits down with it. In the reported case, split 0 of 10 failed and the run produced nothing at all.

Failed splits are now logged and skipped. `ModelFailedException` is raised only when *every* split failed, so an empty backtest is never silently persisted. Training failures remain fatal, since there is no predictor to fall back on.

Only `ModelFailedException` and `NoPredictionsError` are caught, so genuine bugs still propagate.

## Tests

- `test_run_command_failure_reports_output_as_text` -- stdout and stderr appear as text, and the byte-wise stringified form does not.
- `test_backtest_skips_failing_split_and_keeps_the_rest`
- `test_backtest_raises_when_every_split_fails`
- `test_backtest_does_not_swallow_training_failures`

`make lint` and `make test` pass (1244 passed, 117 skipped).

## Related

- [CLIM-930](https://dhis2.atlassian.net/browse/CLIM-930) -- this PR
- [CLIM-929](https://dhis2.atlassian.net/browse/CLIM-929) -- the EWARS model bug that surfaced both defects


knutdrand:
Split into two independent PRs, since the error-reporting fix and the per-split resilience fix are unrelated changes:

- #536 -- `fix(runners): report command output as text on failure`
- #537 -- `fix(assessment): keep usable splits when a backtest split fails`

The combined diff of the two is identical to this one (221 insertions, 39 deletions). Closing in favour of those.

PR #533, opened 2026-08-24
DO NOT MERGE feat: add a WHO-style percentile threshold strategy

mortenoh:
Adds a WHO-style percentile threshold alongside the existing mean + k*std one, so the modeling app's new strategy dropdown has something that matches how malaria surveillance actually sets an endemic channel.

## The problem

Uganda testing showed a district actual around 2,500 against an endemic threshold around 20,000. The cause is that `seasonal` — our only strategy — computes `mean + k*std` over **all** history. Malaria counts are heavily right-skewed and past epidemic years inflate `std`.

Reproduced with five years of one district-month plus one epidemic year:

| | clean history | with one epidemic year |
|---|---|---|
| `mean + 2*std` (ships today) | 2,705 | **28,239** |
| 75th percentile (this PR) | 2,400 | **2,600** |

Actual case level is ~2,500. That is the reported symptom, and the percentile is what fixes it. (The percentile figures here use a 5-year window; the shipped default is 3 — see choice 2.)

Worth noting: a 5-year window **alone does not fix it** — it makes `mean + 2*std` worse (20,567 vs 16,211 in my first check), because the epidemic year becomes a larger share of a smaller sample. The percentile is the fix; the window is a secondary refinement.

## Choices made — please sanity-check these, they may need adjusting

Flagging these explicitly because the right values depend on the data they are applied to:

1. **Default percentile is the 75th (Q3).** The WHO channel's upper/alert line. Overridable with `params: {"quantile": ...}`.
2. **Default lookback is 3 years.** WHO channel practice and the district workbook use 5; the shorter default is a demo preference. Overridable with `params: {"lookback_years": ...}`; `null` uses all history. Zero or negative values are rejected with a 400. Note that quartile robustness to a single outlier needs about five samples: with a 3-year window, one epidemic year inside it drags Q3 halfway to the epidemic level (the interpolated Q3 of three values is the midpoint of the top two).
3. **The window is anchored on the latest requested period**, not on the dataset's newest observation, and is shared by all requested periods.
4. **The anchor year itself is excluded** — asking for 2026 uses 2021–2025. WHO builds the current year's channel from previous *complete* years; including the current year would let an outbreak in progress raise its own threshold. To be precise: the excluded year is the **latest** requested period's year, so a request spanning a year boundary (Dec 2025 + Jan 2026) still has 2025 inside the Dec 2025 baseline. When forecasting this is harmless — the same-season own-year value cannot exist in the data yet. Easy to flip if you'd rather include partial current-year data.
5. **One value per point, not a lower/upper band.** `ThresholdEntry` is `{period, location, value}` and the app's `EndemicThresholdPoint` is `{period, value}` — one line. The lower channel line is reachable today as a second call with `quantile: 0.25`, whenever the chart can draw a band.
6. **`seasonal` is untouched.** Verified byte-identical to master for monthly and weekly data across k = 1.0 / 2.0 / 3.5. It is also called directly by `outbreak_detection.py`, so changing it would have moved metric behaviour.

## The params/UI gap — for later, not now

The modeling app never sends `params`. There is exactly one call site (`useEndemicThresholds.ts:34`) and it passes only `datasetId`, `periodIds`, `strategy`, `locations`.

So `seasonal` has always run at its hardcoded default **`k=2.0`** (`seasonal.py:76`), with no way to tune it from the UI — part of why Uganda's number could not be dialled back. The new strategy will likewise always run at `quantile=0.75` / `lookback_years=5` from the app until a params input is wired up.

That is why this is exposed as a **strategy** rather than a param: the dropdown is the only lever the UI has. Nothing here blocks the dropdown work — `GET /v1/analytics/thresholds/strategies` already returns `id`/`displayName`/`description` and now lists both entries.

The flip side: `useEndemicThresholds.ts` defaults to `strategy = 'seasonal'`, so merging this PR changes nothing visible in the app on its own — it keeps requesting the old strategy until the chap-frontend dropdown (or that default) selects `percentile`. The frontend selection is the critical path for anything users see.

## Changes

- **`chap_core/assessment/thresholds/percentile.py`** (new) — `compute_percentile_thresholds` plus the registered `PercentileThresholdStrategy`, mirroring `seasonal.py`'s shape. Validates `quantile` and `lookback_years`.
- **`chap_core/time_period/vectorized.py`** (new) — the month/week helpers `seasonal.py` kept private, moved so both strategies and the outbreak metrics share them rather than reaching across modules for private names (CLAUDE.md rule 13). They are not threshold-specific, so they live with the rest of the period code instead of under `assessment/thresholds` (not re-exported from the package `__init__` to keep CLI startup imports lean). Implementations are verbatim. Adds `extract_year` for the lookback, one vectorised `^(\d{4})` pattern that covers every supported format (`2016-01`, `2016W01`, `2016-W01`, `2016SunW01`, `2016-S01`, `2016W1`).
- **`outbreak_detection.py`** — import updated to the new module; this also removes its existing `_extract_month` private-name import.
- **`analytics.py`** — the thresholds endpoint maps strategy `ValueError`s to HTTP 400 instead of leaking them as 500. Newly reachable cases: requested periods beyond the data plus lookback (empty window) and an out-of-range `quantile`; also covers the pre-existing frequency-mismatch case.
- **`__init__.py`** — one line in `_discover_strategies()`. No generated-ID lock file exists for thresholds, so `make regen-plot-help` does not apply.

## Tests

Fourteen new cases: thirteen in `test_threshold_strategies.py` — including the epidemic regression above, the lookback window, anchor-year exclusion, quantile ordering, param validation, and a single observation yielding a real value where `mean + k*std` gives `NaN` — plus an endpoint test in `test_from_seeded_engine.py` that an out-of-window request returns 400, not 500.

One fixture added to `conftest.py`. The existing `dataset_observations` derives values from `sin(t % 12)`, which repeats the same value for a given month in **every** year — its within-season variance is exactly zero, so it structurally cannot exercise a percentile.

`make lint` clean; `make test` 1255 passed, 117 skipped.

## Known limitation, not addressed here

`chap_core/assessment/metrics/outbreak_detection.py` calls `compute_seasonal_thresholds` **directly** rather than resolving a strategy from the registry. So the outbreak-detection metrics remain hardwired to `mean + k*std` over unbounded history and do not pick up this strategy or any future one.

That means the inflated thresholds described above still apply wherever those metrics are reported, even after this PR. Routing them through the registry needs a decision on how a metric selects its strategy, so it is deliberately left out. Noted in a comment at the call site.


mortenoh:
close after https://github.com/dhis2-chap/chap-core/pull/539 is merged

PR #394, opened 2026-05-31
Ensemble stacking integration in chap_core

behdadnikkhah25:
# Fix deterministic stacking, NNLS weighting, and sample collapsing

## Summary

This PR finishes the ensemble redesign and addresses the main review comments by:

- making the **deterministic ensemble a clean point-only stacking model** (no residual bootstrap),
- fixing **NNLS usage** so raw coefficients are used for prediction and only normalized for reporting,
- making **probabilistic → deterministic collapsing** explicit (median + warnings),
- documenting and testing the **vincentization** scheme for probabilistic ensembles, and
- improving logging and tests around data issues and fallbacks.

It also:

- removes legacy wrappers in favor of `wrappers.py`,
- moves ensemble config YAMLs to `tests/fixtures/ensemble_config/`,
- removes the unused `use_residual_bootstrap` flag from models and CLI, and
- keeps all tests passing (`964 passed, 117 skipped, 4 xfailed, 1 xpassed`).

---

### Deterministic ensemble = clean stacking (no bootstrap)

- Residual bootstrap has been **entirely removed** from:
  - `EnsembleModel` / `EnsembleEstimator`, and
  - the ensemble CLI.
- Deterministic ensembles now:
  - output a **single sample** per forecast (`sample_0`),
  - log a warning that **CRPS reduces to MAE** and is not directly comparable to multi-sample CRPS.
- `use_residual_bootstrap` is **removed** from:
  - `EnsembleModel`,
  - `EnsembleEstimator`, and
  - CLI signatures and tests.

### NNLS coefficients: prediction vs reporting

- `NonNegativeMetaModel.fit` stores the raw NNLS solution in `coef_`.
- `NonNegativeMetaModel.predict` uses `X @ coef_` directly (no renormalization).
- `EnsembleModel.train`:
  - reads `coef_`, clips to non-negative,
  - normalizes **only for logging/reporting** as percentages in `self.weights`.
- This avoids the previous “optimize one objective, predict with another” behavior.

### Probabilistic → deterministic collapse (median + warnings)

- When probabilistic base models are used in the deterministic ensemble:
  - `SampleExtractor.samples_to_flat` detects `sample_*` columns,
  - logs a warning that **uncertainty is discarded**,
  - collapses to a point forecast via the **median** across samples per row.
- This replaces the previous mean-based, silent collapse.

### Probabilistic meta-model: vincentization

- `ProbabilisticMetaModel` now combines base distributions via **vincentization (quantile averaging)**:
[...]

PR #534, opened 2026-08-25
fix: rebase ensemble stacking onto master and fix review findings

knutdrand:
Supersedes #394. Same ensemble stacking work by @behdadnikkhah25, with that
branch's history preserved, merged onto current master and with the review
findings fixed on top.

Opened against upstream rather than into the fork branch so that CI actually
runs. #394 is opened from a fork's `master`, so the workflows never triggered
and the CRPS breakage below went unnoticed for three months.

Everything specific to this PR is in the single commit `fix: rebase ensemble
stacking onto master and fix review findings` (14 files, +378/-160). The rest
is the original branch plus the master merge.

## Merge fallout

- **Dropped the branch's CRPS rewrite.** Master vectorised CRPS in #467 after
  this branch was cut, so the performance work was redundant. The branch's
  version also switched the estimator from the biased `1/m^2` form to the fair
  `1/(m(m-1))` form, which broke three lock-in tests and silently changed the
  value of every stored CRPS. Core keeps one definition; a new public
  `crps_matrix` helper gives the ensemble a batch scorer that agrees with
  `CRPSMetric` row for row.
- **Reverted the TEMP PATCH to `cli_endpoints/_common.py`.** It modified core
  code and dropped the CSV header it was meant to restore.

## Correctness

- **Target leakage.** `disease_cases` was reaching base models during inner
  validation. `ExternalModel.predict` writes `future_data` verbatim to the CSV
  it hands the model, so base models could read the very values the
  meta-weights are fitted against. The target is now masked, matching what
  `dataset_splitting` already does everywhere else.
- **Horizon mismatch.** Weights were fitted on one 12-period window but applied
  at `n_periods` (default 3). Relative model skill is strongly horizon
  dependent, so this ranked base models at the wrong horizon. Inner validation
  now runs as rolling windows of `n_periods`, reusing the trained base models
  with expanding history so there is no retraining cost.
- **Chapkit path was unusable.** Templates were never entered as context
  managers, so `--run-config.is-chapkit-model` raised `RuntimeError` and
  nothing shut the service down. Now handled with an `ExitStack`.
- **Zero-weight fallback.** The uniform fallback only touched the reported
  weights, leaving `coef_` at zero, so the ensemble forecast zeros while
  logging an equal-weight blend. Both meta-models now fall back internally.
- **Resampling.** One index set was drawn with replacement and reused for every
  row, distorting all locations' spread identically. Each row's empirical
  quantile function is interpolated instead.
- Guarded the probabilistic branch against missing base-model rows, and added a
  row-count check before falling back to row-order alignment.
- Dataset URLs resolve via `resolve_csv_path` and the dataset is validated
  before backtesting, matching `chap evaluate`.
- Removed the unreachable 0-based `horizon_distance` filter; the column is
  1-based. The CRPS-equals-MAE warning fires once instead of once per split.

## API change

The resampling fix makes the ensemble deterministic, so `--random-state` no
longer did anything. It is replaced by `--inner-val-periods`, which exposes the
inner validation length that was previously hardcoded to 12.

## Verification

`make lint` clean (ruff, mypy on 415 files, pyright 0 errors).
`make test`: 1292 passed, 121 skipped, 4 xfailed, 1 xpassed.

For reference, merging #394 as-is onto current master fails 19 tests in
`tests/evaluation/test_metrics.py`.


knutdrand:
## Interference audit

Checked what this PR touches outside `chap_core/ensemble/`, and fixed two things it was doing to the rest of chap.

### Fixed: CLI startup regression

Registering `evaluate-ensemble` in `cli.py` imported the endpoint eagerly, which pulled scipy, chapkit, fastapi, starlette and the DB layer into **every** `chap` invocation. Measured in one venv:

| | startup | scipy | chapkit |
|---|---|---|---|
| without ensemble endpoint | 0.59s | no | no |
| with, before fix | 1.26s | yes | yes |
| with, after fix | 0.59s | no | no |

The heavy imports now happen inside the command body, matching what `evaluate.py` already does. Zero measurable cost.

### Fixed: unrelated changes removed

- `tests/evaluation/test_evaluation_serialization.py` had a numpy ABI warning workaround that suppressed a `RuntimeWarning` in six existing tests, plus two new tests for that warning. Nothing to do with ensembles; reverted to master, suite still green.
- `pyproject.toml` added mypy and pyright excludes for `chap_core/runs`. `CHAP_RUNS_DIR` defaults to `runs/`, which master already excludes, so this was an artifact of running the CLI from inside the package. Removed; both type checkers still pass.

### Remaining surface outside the ensemble package

**+66 lines, 0 deletions**, all purely additive:

| file | change |
|---|---|
| `assessment/metrics/crps.py` | adds `crps_matrix`; `_crps_sample`, `CRPSMetric`, `CRPSLog1pMetric` untouched, so no existing metric value changes |
| `cli.py` | 2 lines registering the command |
| `tests/evaluation/conftest.py` | one new fixture |
| `tests/evaluation/test_metrics.py` | three new tests for `crps_matrix` |

`cli_endpoints/_common.py` is untouched, which also clears the standing "no changes in core code" review thread from #325.

`make lint` clean. `make test`: 1290 passed, 121 skipped, 4 xfailed, 1 xpassed.

knutdrand:
## Live validation against real external models

Ran `evaluate-ensemble` end-to-end on LAO admin1 monthly (18 regions, 156 periods) with two real `ExternalModel` base models that go through the full MLflow subprocess and CSV round-trip. Both modes exit 0.

**The leakage fix is confirmed on real generated files.** The CSVs actually handed to the base models:

```
future_data_2010-03-01.csv:   ,time_period,population,location_name,rainfall,mean_temperature,mean_relative_humidity,parent,location
historic_data_2010-03-01.csv: ,time_period,disease_cases,population,location_name,rainfall,...
```

No `disease_cases` in future data; still present in historic. `2010-03-01` and `2010-06-01` are the two inner-validation windows written before the `Meta-weights` log line, which are exactly the ones that were leaking.

Other fixes confirmed live:

- `Inner validation: 152 periods, train=146, val=6, 2 window(s) of horizon 3` — rolling-origin windows at the backtest horizon
- `Rejected regions: ['LA-VI'] due to missing target values` — the added dataset validation firing
- report CSV header restored: `Model,ratio_above_truth,crps,crps_log1p,...`
- the CRPS-equals-MAE warning fires once, not once per split

Sanity signals: deterministic run gives `crps == mae` exactly (121.3933299242252), matching what the warning claims. Probabilistic run gives `crps` 97.22 below `mae` 115.25, so vincentization is producing a real predictive distribution rather than a collapsed point forecast.

### Not covered by this run

- Weights came out as corner solutions (`[100, 0]` deterministic, `[0, 100]` probabilistic) because the two available offline base models are near-identical and therefore collinear. A genuinely blended weight vector was not exercised.
- The chapkit path is still only covered by unit test, not live: no docker in this environment, so the `with template:` fix has not been exercised against a real chapkit service.
- MLflow's file-store backend is broken in this environment (`Could not find experiment with ID 0`), but that is pre-existing and unrelated: master's own `chap eval` fails identically. These runs used `MLFLOW_TRACKING_URI=sqlite:///...`.

behdadnikkhah25:
I’ve pulled the branch locally and run `evaluate-ensemble` end-to-end on the LAO admin1 monthly dataset with three external base models (`rwanda_sarimax`, `chap_auto_ewars`, `INLA_baseline`), using the script I shared. Both probabilistic and deterministic modes exit 0. I also ran each base model individually with the same backtest configuration and compared everything via `chap export-metrics`.

## Probabilistic ensemble (LAO admin1 monthly, 3 splits, horizon 3)

Inner validation: 151 periods, rolling-origin windows at `n_periods=3`, `n_splits=3`, `stride=1`, dropping rows with NaNs in targets/features.

Meta-weights:
- `rwanda_sarimax`: ≈11.5%
- `chap_ewars_monthly`: ≈2.9%
- `INLA_baseline`: ≈85.6%

Metrics:
- CRPS ≈ 83.14, MAE ≈ 117.30, RMSE ≈ 247.31
- Coverage 10–90 ≈ 0.71, Coverage 25–75 ≈ 0.49
- `sample_count` = 14 400 (100 samples per forecast)

CRPS is below MAE, and coverage/Winkler scores behave sensibly for a genuine predictive distribution.

## Deterministic ensemble

Base-model samples are collapsed to a single point forecast (median), as indicated by the “Collapsing probabilistic samples…” warnings.

Meta-weights:
- `rwanda_sarimax`: ≈75.5%
- `chap_ewars_monthly`: 0.0%
- `INLA_baseline`: ≈24.5%

Metrics:
- CRPS = MAE ≈ 177.62
- RMSE ≈ 291.15
- Coverage 10–90 = 0, Coverage 25–75 = 0
- `sample_count` = 144 (one sample per forecast)

The equality CRPS = MAE matches the “CRPS reduces to MAE” warning and confirms that the deterministic ensemble is behaving as a pure point-forecast meta-model.

## Base models (same dataset and backtest config)

From `metrics_comparison_25_08.csv`:
- `rwanda_sarimax`: CRPS ≈ 196.59, RMSE ≈ 324.27
- `chap_auto_ewars`: CRPS ≈ 91.94, RMSE ≈ 304.33
- `INLA_baseline`: CRPS ≈ 83.83, RMSE ≈ 269.19

These line up well with the ensemble weights: the probabilistic ensemble leans on INLA (best CRPS), while the deterministic ensemble leans on `rwanda_sarimax` (better point forecasts in this setup).

## Checklist against your earlier comments

As far as I can see, all the points you asked me to verify have now been checked:

- End-to-end runs of `evaluate-ensemble` in both probabilistic and deterministic modes on a real dataset (LAO admin1 monthly) with multiple external base models.
- Confirmation that the probabilistic meta-model uses rolling-origin inner validation at the backtest horizon, drops rows with NaNs, and produces reasonable weights.
- Verification that in deterministic mode we collapse samples to a point forecast, CRPS equals MAE, and the corresponding warnings are emitted.
- Separate evaluation of each base model with the same backtest settings and comparison via `export-metrics`.
- Check that `disease_cases` is not present in the future-data CSVs used during inner validation (fixing the target-leakage issue).
- Sanity around region filtering (LA-VI is rejected due to missing targets over the training period).

If there is anything else from your earlier messages that you feel is not yet covered, I can run additional checks.

## Inner validation design

On the inner validation design: your original idea of a single 12-period inner window is reasonable, but using rolling windows at the backtest horizon (`n_periods = horizon`) feels more consistent with how the ensemble is actually evaluated in CHAP. We validate the weights on the same type of forecasts (same horizon and rolling-origin structure), and the logs (`train=139, val=12, 4 windows of horizon 3`) plus the resulting meta-weights reflect that.

## Review position / merge readiness

From my side, the implementation behaves as intended and matches the architecture/design I’ve been working towards. The end-to-end evaluation on LAO with three base models is consistent with the expected behaviour of both the probabilistic and deterministic ensembles, and I don’t see any remaining changes I need to make on this branch based on my testing. I would consider this PR ready to merge.

mortenoh:
Reviewed the full diff against master (22 files, +1766) and ran the new suites (`tests/ensemble` + `tests/cli_endpoints/test_ensemble_cli.py`) in a throwaway worktree - 26 passed.

The listed correctness fixes hold up under reading: the target masking matches `dataset_splitting.remove_field("disease_cases")`, `crps_matrix` is byte-for-byte the same estimator as `_crps_sample` (`1/n^2`, order-statistic form), the `ExitStack` genuinely spans the whole backtest (`Evaluation.from_samples_with_truth` fully consumes the `backtest` generator before returning), and both meta-models now fall back internally rather than only in the reported weights.

## Medium

**1. `chap_core/cli_endpoints/ensemble.py:215` - missing base-model prediction-length handling.** `evaluate-ensemble` skips the checks `chap evaluate` applies (`evaluate.py:242-259`): no `min_prediction_length` error, no `ExtendedPredictor` wrapping when `max_prediction_length < n_periods`, and no `warn_unused_covariates`. A base model capped at 1 period, run with the default `--backtest-params.n-periods 3`, is silently asked for 3; the short prediction frame then merges to NaN and blows up mid-backtest with `Missing base model predictions for N of M rows` (predict) or drops the rows from the weight fit (train), instead of being wrapped as it would be under `chap evaluate`.

**2. `chap_core/ensemble/ensemble_model.py:92` - trailing inner-validation window can be shorter than `horizon`.** `range(split_idx, len(periods), self.horizon)` with `--inner-val-periods 10 --backtest-params.n-periods 4` yields windows of 4, 4, **2**. The tail window scores base models at horizons 1-2, and its rows are concatenated into the same CRPS/NNLS fit as the 4-step rows - partially reintroducing the horizon mismatch this PR fixes. The existing test only covers the evenly divisible case (6 / 3). Either drop a trailing partial window or offset `split_idx` so the remainder divides.

**3. `chap_core/ensemble/_sample_extractor.py:83` - unreachable fallback with a misleading error.** This branch is entered only when `df_pred` lacks `location`/`time_period`, but it then calls `samples_to_flat`, which raises `ValueError(f"Missing {missing} in prediction DataFrame")` for exactly those columns, and then merges on them. Verified: a predictor returning `{time_period, forecast}` (no location, no `sample_*`) fails with `Missing ['location'] in prediction DataFrame` rather than taking the documented "repeat point forecasts" path. Either drop the branch or align on row order like the sibling `if sample_cols:` branch above it.

## Low/medium

**4. `chap_core/ensemble/ensemble_model.py:201` - reported weights do not describe the actual combination for `method="deterministic"`.** `NonNegativeMetaModel.predict` applies the raw NNLS coefficients (`test_non_negative_meta_model_keeps_nnls_scale` locks this in), whose sum need not be 1, while `self.weights` - used for the log lines and `ensemble_meta_report.csv` - is normalised to 100 %. NNLS returning `[0.4, 0.4]` is reported as 50 %/50 %, implying a plain average, but the forecast is `0.8 x average`. The normalisation discards exactly the shrinkage information a reader needs.

## Low

- **`ensemble_model.py:181`** - `self.meta_model` is cached on the estimator and re-fitted in place on every `train()`. With `--backtest-params.n-retrain 2` (a valid `BacktestParams` field the command accepts) `train` runs twice, so `ensemble.weights` read at `cli_endpoints/ensemble.py:245` and the meta report describe only the final retrain, with nothing in the file saying so. It also mutates the `_meta` of an `EnsemblePredictor` handed out by an earlier `train()`.
- **`cli_endpoints/ensemble.py:110`** - `_write_meta_report` always writes `ensemble_meta_report.csv` next to `report_filename`, ignoring the report's stem. Two runs in the same directory with different `--report-filename` produce distinct `.csv`/`.0.csv`/`.nc` outputs but silently clobber each other's weight report. `report_filename.with_name(report_filename.stem + "_meta.csv")` would fix it.
- **`cli_endpoints/ensemble.py:221`** - `n_samples=100` is hard-coded with no flag. A base model producing e.g. 10 samples is silently upsampled to 100 by `_resample_to_quantiles`, presenting a distributional resolution the base model never had, with no way to lower it.
- **`cli_endpoints/ensemble.py:160`** - `ensemble_method` is validated after `_load_dataset` (which may `pooch.retrieve` a remote CSV plus companion GeoJSON) and after `validate_and_filter_dataset_for_evaluation`, so a typo in `--ensemble-method` costs a full download before failing. Move the check above the dataset load.
- **`ensemble_model.py:51`** - `raise ValueError(method)` produces a bare `ValueError: probabalistic` with no indication of what is wrong or what the valid values are; the neighbouring guards all carry proper messages. `test_invalid_method_raises` matches on the echoed value, so it would keep passing with a real message.
- **`tests/fixtures/ensemble_config/*.yaml`** - all three fixture YAMLs are added but referenced by nothing; `grep -rn ensemble_config` over `chap_core/` and `tests/` returns no hits. Dead files carried over from the original branch.
- **`_meta_models.py:15`** - `crps_ensemble` is a bare one-line pass-through to `crps_matrix`, exported from both `_meta_models.__all__` and `ensemble_model.__all__`, and called from nowhere. Same for `intercept_` on both meta-models (lines 35 and 71): set to `0.0`, never read by either `predict`.
- **`cli_endpoints/evaluate.py:238`** - pre-existing, but now inconsistent: `chap evaluate --estimator-options.mode ensemble` still raises `NotImplementedError("Ensemble mode is not yet implemented")`. With `evaluate-ensemble` landing as a separate command there are now two ensemble entry points, one permanently dead. Worth either wiring `EstimatorMode.ENSEMBLE` to the new path or naming `evaluate-ensemble` in the error.

## Checked and cleared

Row alignment across the train/predict paths is consistent - `df_val`, the masked windows, and every `merge` key off `future.to_pandas()` order, and `_pack_samples`/the deterministic reindex both index positionally within a location, so the unmasked-target frame and the masked feature frames stay in lockstep. Partial-NaN rows surviving `_resample_to_quantiles` with distorted quantiles are always either dropped (train) or hard-failed (predict), so the distortion never reaches a forecast. `create_model_lists` guarantees equal-length name/config lists, so the `strict=False` zip at `ensemble.py:194` and the meta-report header/row widths cannot desync. Mutable pydantic defaults on `backtest_params`/`run_config` match the existing pattern in `evaluate.py`, `report.py`, and `explain.py`. The new module keeps scipy/chapkit/DB imports inside function bodies, so `tests/test_command_line_interface_adaptor.py`'s startup guard stays green.

knutdrand:
Thanks @mortenoh - all of these are fixed in 7c74644d. Point by point:

## Medium

**1. Missing base-model prediction-length handling.** `evaluate-ensemble` now runs the same checks `chap evaluate` does, per base model, before the backtest starts: `warn_unused_covariates`, a hard error when `min_prediction_length > n_periods`, and `ExtendedPredictor` wrapping when `max_prediction_length < n_periods`. The wrapping lives in `TemplateWithConfig` via a new `extend_to_prediction_length`, so both the inner-validation estimators and the full-data estimators get it. Covered by `test_base_model_below_minimum_prediction_length_is_rejected` and `test_base_model_above_maximum_prediction_length_is_extended` - the latter runs the CLI with `n_periods=2` against a base model capped at 1 and asserts the model is never handed more than one period.

**2. Trailing inner-validation window shorter than `horizon`.** Windows are now whole-horizon only. `n_windows = (len(periods) - split_idx) // horizon`, and the remainder is moved back into the inner training data rather than forming a short tail window; `--inner-val-periods 10 --backtest-params.n-periods 4` now yields two windows of 4 instead of 4, 4, 2. If not even one full window fits, it raises with the two knobs to change instead of silently fitting on the wrong horizon. `test_inner_validation_drops_partial_trailing_window` covers the non-divisible case you pointed out.

**3. Unreachable fallback in `_sample_extractor.py`.** You were right that the branch could only raise. It now reads `forecast`/`value` directly off `df_pred` instead of going through `samples_to_flat`, and applies the same row-count check as the sibling `if sample_cols:` branch before aligning by row order. A predictor returning `{time_period, forecast}` now takes the documented repeat-point-forecasts path. Two tests, one for the happy path and one for the length mismatch.

## Low/medium

**4. Reported weights vs. the actual combination.** The meta report now carries both quantities, one row each, per fit:

```
Model,round,quantity,model_a,model_b
ensemble_meta,1,weight_percent,50.000000,50.000000
ensemble_meta,1,coefficient,0.400000,0.400000
```

so the NNLS shrinkage is visible rather than normalised away. The per-model log line gained the raw coefficient, and deterministic runs whose coefficients do not sum to 1 log the scale factor explicitly.

## Low

- **`n_retrain`.** `train()` builds its own meta-model instead of re-fitting a cached one, so an `EnsemblePredictor` handed out by an earlier `train()` is no longer mutated. Every fit is appended to `fit_history` and every round gets its own rows in the report, so nothing is silently last-round-only.
- **Meta report filename.** Now `report_filename.with_name(stem + "_meta.csv")`. Test asserts `run_one.csv` produces `run_one_meta.csv`.
- **`n_samples`.** Exposed as `--n-samples` (default 100), with help text saying that upsampling cannot add resolution the base model does not have.
- **`ensemble_method` validated late.** Moved above `_load_dataset`, so a typo fails before any download.
- **`raise ValueError(method)`.** Now `method must be 'deterministic' or 'probabilistic', got 'probabalistic'`. `test_invalid_method_raises` still matches on the echoed value.
- **`tests/fixtures/ensemble_config/*.yaml`.** Deleted.
- **`crps_ensemble` and `intercept_`.** Removed, along with the `__all__` entries.
- **`evaluate.py:238`.** The error now reads "Ensemble mode is not implemented for `chap evaluate`; use the `chap evaluate-ensemble` command". Wiring `EstimatorMode.ENSEMBLE` to the new path is a bigger change than this PR should carry while the command is still marked experimental.

`make lint` clean (ruff, mypy on 415 files, pyright 0 errors). `make test`: 1302 passed, 121 skipped, 4 xfailed, 1 xpassed (10 new tests).

Not re-run live: the end-to-end LAO runs above predate these changes. The prediction-length wrapping in particular is only covered by the CLI test, not against a real capped model.

mortenoh (approved):

PR #536, opened 2026-08-26
fix(runners): report command output as text on failure

knutdrand:

Split out of #528, which bundled this with an unrelated backtest resilience change. The other half is in the companion PR linked below.

## The problem

`chap_core/runners/command_line_runner.py` called `process.communicate()` a second time after the first call had already consumed the streams, then embedded the result with `''.join(map(str, streamdata))`. Iterating `bytes` yields `int`, so each byte was rendered as a decimal number:

```
>>> out, err = p.communicate()
>>> ''.join(map(str, p.communicate()[0]))
'10410510'          # this is b'hi\n'
```

In a reported EWARS failure this produced a 12,167-digit run of concatenated ASCII codes carrying no information that the decoded output did not already contain. Combined with the exception being re-wrapped into `ModelFailedException` and printed in both halves of the `__cause__` chain, the R log appeared four times in what should have been a ~50 line report.

## The change

- Drop the redundant `communicate()` call; build the message once from the already decoded stdout/stderr.
- Decode with `errors="replace"`. Model output is not guaranteed to be valid UTF-8 (locale-dependent R warnings, for instance), and a failed model must still produce a readable error rather than a `UnicodeDecodeError`.
- Remove the commented-out streaming block and the `except subprocess.CalledProcessError` handler, which could never fire since `Popen` does not raise it.

## Tests

- `test_run_command_failure_reports_output_as_text` -- stdout and stderr appear as text, and the byte-wise stringified form does not.
- `test_run_command_failure_with_non_utf8_output_is_still_readable`

`make lint` clean; `tests/runners/test_runners.py` 25 passed, 1 skipped.

## Related

- Supersedes half of #528
- [CLIM-929](https://dhis2.atlassian.net/browse/CLIM-929) -- the EWARS model bug that surfaced this

### DIFF
--- a/chap_core/runners/command_line_runner.py
+++ b/chap_core/runners/command_line_runner.py
@@ -33,41 +33,23 @@ def run_command(command: str, working_directory=Path("."), env: dict | None = No
         Environment variables to use. If None, uses the current environment.
     """
     logging.debug(f"Running command: {command}")
-    # command = command.split()
+    process = subprocess.Popen(
+        command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, cwd=working_directory, shell=True, env=env
+    )
+    stdout, stderr = process.communicate()
+    # Model output is not guaranteed to be valid UTF-8 (locale-dependent R
+    # warnings, for instance); a failed model must still produce a readable
+    # error message rather than a UnicodeDecodeError.
+    output = stdout.decode(errors="replace") + "\n" + stderr.decode(errors="replace")
+    return_code = process.returncode
 
-    try:
-        process = subprocess.Popen(
-            command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, cwd=working_directory, shell=True, env=env
+    if return_code != 0:
+        message = (
+            f"Command '{command}' failed with return code {return_code}, "
+            f"Full output from command below: \n ----- \n{output} \n--------"
         )
-        stdout, stderr = process.communicate()
-        output = stdout.decode() + "\n" + stderr.decode()
-        """
-        output = []
-        for c in iter(lambda: process.stdout.read(1), b""):
-            sys.stdout.buffer.write(c)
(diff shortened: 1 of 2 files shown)


mortenoh:
Reviewed. The core change is sound: the duplicate `process.communicate()` is gone, the failure message is built once from the already-decoded streams, `errors="replace"` keeps a `UnicodeDecodeError` from masking the real model failure, and the removed `except subprocess.CalledProcessError` handler was dead code (`Popen` never raises it). No caller depends on the old message format - checked every `except CommandLineException` site in `external_model.py` plus the renv/uv/conda/docker runners.

One issue found and fixed in c8826488:

**`tests/runners/test_runners.py` - the non-UTF-8 regression test was vacuous on CI.** `subprocess(shell=True)` runs `/bin/sh`, which is dash on `ubuntu-latest`, and dash's `printf` does not implement hex escapes. `/bin/dash -c "printf 'boom-\xff-stderr'"` emits the literal ASCII text `boom-\xff-stderr`, so no `0xff` byte was ever produced and `decode()` never saw invalid UTF-8. Both assertions passed either way, meaning reverting `errors="replace"` would not have failed CI. Switched to the POSIX octal escape (`\377`, works under both dash and bash) and added an assertion on the U+FFFD replacement character. Verified by mutation: with `errors="replace"` removed, the test now fails with `UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 5`.

One non-blocking note: `logger.error(message)` in `command_line_runner.py` logs the full model output, and the same string is carried by the raised `CommandLineException`, which `external_model.py` re-wraps as `ModelFailedException(str(e)) from e` and logs again via the traceback. For the EWARS-scale R logs this PR targets, the whole log lands in the stream at least twice. Consider logging a one-line summary (command + return code) and letting the exception carry the output.

mortenoh (approved):

PR #537, opened 2026-08-26
fix(assessment): keep usable splits when a backtest split fails

knutdrand:

Split out of #528, which bundled this with an unrelated error-reporting fix. The other half is #536.

## The problem

`chap_core/assessment/prediction_evaluator.py:backtest` let a per-split `predict` exception escape the generator. A model that fails on a single split -- numerical instability, a degenerate window -- took all the other splits down with it. In a reported EWARS backtest (9 provinces, 10 years, 5 covariates), split 0 of 10 failed and the run produced nothing at all.

The model bug itself -- a rank-deficient dlnm lag basis at `n_lags = 3` -- lives in the model repos and is not addressed here; it is tracked in [CLIM-929](https://dhis2.atlassian.net/browse/CLIM-929).

## The change

- Failed splits are logged and skipped instead of aborting the run.
- While no split has yet succeeded, two consecutive failures abort the backtest, so a model that is broken for the whole dataset does not pay for a full model run per remaining split.
- A run where every split failed raises `ModelFailedException`, so an empty backtest is never silently persisted.
- Training failures remain fatal; there is no predictor to fall back on.
- `run_command_through_docker_container` raised a bare `AssertionError` on a non-zero container exit. It now raises `CommandLineException` like `CommandLineRunner`, so docker-backed model failures reach the same per-split handling, and its logs decode with `errors="replace"`.

## Tests

- `test_backtest_skips_failing_split_and_keeps_the_rest`
- `test_backtest_skips_raw_exception_split_and_keeps_the_rest`
- `test_backtest_tolerates_one_initial_failure`
- `test_backtest_aborts_early_when_no_split_succeeds`
- `test_backtest_does_not_swallow_training_failures`
- `test_failed_container_raises_command_line_exception`

ruff and mypy clean; the two touched test modules pass (11 passed, 3 skipped).

## Related

- Supersedes half of #528
- #536 -- the error-reporting half
- [CLIM-929](https://dhis2.atlassian.net/browse/CLIM-929) -- the EWARS model bug that surfaced this

### DIFF
--- a/chap_core/assessment/prediction_evaluator.py
+++ b/chap_core/assessment/prediction_evaluator.py
@@ -24,11 +24,19 @@ from chap_core.assessment.dataset_splitting import (
 )
 from chap_core.data.gluonts_adaptor.dataset import ForecastAdaptor
 from chap_core.datatypes import Samples, SamplesWithTruth, TimeSeriesData
+from chap_core.exceptions import ModelFailedException
+from chap_core.log_config import get_status_logger
 from chap_core.spatio_temporal_data.temporal_dataclass import DataSet
 from chap_core.time_period import PeriodRange
 
 plt.set_loglevel(level="WARNING")
 logger = logging.getLogger(__name__)
+status_logger = get_status_logger()
+
+# With no successful split yet, this many consecutive failures is treated as a
+# systemically broken model and aborts the backtest instead of retrying every
+# remaining split.
+_MAX_INITIAL_FAILURES = 2
 
 
 FeatureType = TypeVar("FeatureType", bound=TimeSeriesData)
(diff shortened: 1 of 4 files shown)


mortenoh:
Reviewed. The core change is sound: `predictor.predict` wrapped in a broad `except`, training left outside the `try` so training failures stay fatal, fail-fast before the first success, and a terminal `ModelFailedException` when nothing succeeded. The `CommandLineException` swap in `docker_helper_functions.py` correctly routes docker model failures into `ExternalModel.predict`'s existing `except CommandLineException -> ModelFailedException` handler. The touched test modules pass on a worktree of the PR head (12 passed, 2 skipped).

Four findings, the first two worth fixing before merge.

**1. `chap_core/assessment/prediction_evaluator.py:153` - medium: `merge` sits outside the `try`.** `future_truth.merge(r, result_dataclass=SamplesWithTruth)` is outside the protected block, so a malformed-but-returned prediction still kills the whole backtest - the exact failure mode this PR sets out to contain. `DataSet.merge` (`temporal_dataclass.py:575`) asserts every truth location is present in the prediction, and `TimeSeriesData.merge` (`datatypes.py:195-196`) raises `ValueError` on a period-range mismatch. Scenario: on one degenerate window an R model drops a province it could not fit, or emits a shifted time range; `predict` returns normally, `merge` raises `AssertionError`/`ValueError`, the generator dies, and all nine other splits are discarded - no skip, no `ModelFailedException`. Moving the `merge` inside the same `try` closes this.

**2. `chap_core/docker_helper_functions.py:90` - medium: container leak on the new failure path.** The new `raise CommandLineException(...)` returns before `container.remove()` on line 100, and `auto_remove` is `remove_after_run`, which defaults to `False` (no caller passes `True`; `DockerRunner._execute` uses the default). The old `assert` had the same gap but aborted the entire run, so at most one container leaked per backtest. Now failures are a routine, tolerated path: a 10-split backtest against a docker model that fails on 9 splits leaves 9 stopped containers and their writable layers on the host, and every re-run adds more. `DockerRunner.teardown()` is a no-op, so nothing reclaims them. Wrap lines 87-100 in `try/finally`, or call `container.remove()` before raising.

**3. `chap_core/assessment/prediction_evaluator.py:161` - low/medium: no failure threshold once a split succeeds.** After the first success there is only a `logger.warning`/`status_logger.warning`. If split 0 succeeds and splits 1-9 all fail, `run_backtest` (`db_worker_functions.py:118-146`) persists a Backtest built from one split, computes `aggregate_metrics` over 1/10 of the intended data, and logs "Backtest completed successfully". The API consumer sees a normal successful backtest with no signal that 90% is missing, which sits awkwardly against the PR's own goal that an empty backtest is never silently persisted. Consider a minimum-success fraction, or recording the skipped splits on the Backtest row.

**4. `chap_core/assessment/prediction_evaluator.py:150` - low: `None` returns increment neither counter.** `if r is None: continue` means `n_predicted` is not "splits that ran" and `n_failed + n_predicted` is not the split count, so the summary message reports a wrong denominator and `if not n_predicted` treats None-returning splits as failures-in-waiting. Scenario: `chap evaluate --dry-run` drives models whose `predict` returns `None` for every split; if one split raises, the run ends with `ModelFailedException("All 1 evaluation splits failed")` even though nine dry-run splits completed fine, and two such raises trip the early abort. Counting None-returning splits explicitly, or gating the terminal check on `n_failed == total_splits`, would make the messages truthful.

Checked and clear: training failures correctly stay outside the `try` and `n_retrain > 1` retrains remain fatal as documented; all `backtest()` consumers fully exhaust the generator, so the terminal `if n_failed:` block does run; nothing in the repo catches `AssertionError` from the docker helper, so the exception-type change breaks no caller, and `tests/external/test_docker.py:31`'s `docker.errors.APIError` expectation is raised at container creation and unaffected; module-scope `get_status_logger()` is idempotent; skipped splits leave gaps in `backtest.split_periods`, but downstream consumers use set operations and tolerate them.

knutdrand:
In my opinion we should not merge this one. Should just fail on first error so that it can be resolved and so that we don't hide unstable behavior

PR #538, opened 2026-08-27
docs: point local DHIS2 setup at docker-dhis2-core

hansand02:
Updated the guide to run Modeling app locally with details @mortenoh told me on tuesday:)

Instead of mentioning a test database on an internal drive, point it at the public [docker-dhis2-core](https://github.com/dhis2-chap/docker-dhis2-core)
instead.

- Split the DHIS2 setup into two clearly marked options (demo database vs. a
  plain instance via `d2 cluster up`) <- this confused me a bit when first reading the docs
- Within the docker-dhis2-core option, distinguish letting the stack run
  chap-core (`make start-chap`) from running chap-core yourself (`make start`).
- Add the default test credentials for the demo database.

"make docs" runs fine


hansand02:
Updated this with some new info i got that unblocked me. hope this is the general truth/best for all option!

Edit: Im not able to request reviewers due to access rights @knutdrand . Will this be auto assigned to a reviewer?

mortenoh (commented):
Reviewed the docs change. The factual claims all hold up: the `/health` response body matches `chap_core/rest_api/common_routes.py` exactly and `/health` is in `OPEN_PATHS`, so it works even with `CHAP_API_TOKEN` set; port 8000 is published on the host in both compose setups; `make start` / `make start-chap` exist upstream and the demo dump, analytics run and route registration really are automatic; and `make start` defaults `CHAP_ROUTE_URL` to `host.docker.internal` with a matching `extra_hosts` entry, so option B works on Linux too. Both new links resolve, and no other page referenced the renamed headings.

Five comments on `docs/contributor/local-dhis2-installation.md`, only the first of which I would hold the merge for.

**Line 12 -- "With either option ..." is ambiguous, and the wrong reading breaks the setup.** "option" now means two different things on this page. Here it means the A/B choice inside docker-dhis2-core, but the surrounding headings are "Option 1" and "Option 2", and line 21 uses "Whichever option you chose" to mean 1 vs 2. Because the sentence sits directly above the `### Option 2` heading, a reader following Option 2 (`d2 cluster up`) will reasonably read it as covering their path, skip the CHAP URL step at line 25, and end up with a Modeling app that cannot reach CHAP. Suggest scoping it explicitly, e.g. "With both A and B, docker-dhis2-core downloads and loads the demo database on first start, runs analytics for you, and registers the DHIS2 Route ...".

**Lines 25 and 34 -- the "skip the manual URL step" promise contradicts what follows.** Line 25 still tells everyone to enter the CHAP URL, and line 34 then lists "The Modeling app is configured with the correct URL for chap-core (the previous step)" as a hard prerequisite for running an evaluation. A docker-dhis2-core user who correctly skips the step hits a prerequisite telling them they must have done it. Both spots need an "Option 2 only" / "if the route was not registered for you" qualifier.

**Line 10 -- option B links to the wrong install path.** It says "Choose this if you are working on the chap-core code", but the linked First-time Setup page is the Docker-image install (`git checkout [VERSION]` plus `docker compose -f compose.yml -f compose.chapkit.yml up -d`). That runs a pinned release image, not your working tree, so it does not serve the stated use case -- a contributor following it will start a released image and then wonder why their changes have no effect. Upstream's own README describes this path as a chap-core you run from source on the host, so the link should point at the contributor/source-run instructions instead of (or in addition to) First-time Setup.

**Line 9 -- `make start-chap` pulls more than DHIS2 and chap-core.** Upstream's target runs `docker compose -f compose.chapkit.yml up`, and `compose.chapkit.yml` is the umbrella that pulls in every chapkit model overlay (EWARS and friends) on top of DHIS2 and chap-core. Worth a note, since someone on a slow connection gets several extra model images with no warning.

**Lines 5 and 14 -- heading levels skip a level.** They jump from `#` (line 1) straight to `###`, then back up to `##` on lines 19 and 29. The two Options render as h3 with no h2 parent, so in the mkdocs-material sidebar they sit deeper than the "After DHIS2 is up" section that follows them. Both should be `##`.

hansand02:
Reviewed your comments @mortenoh 

Quite a bit to grasp, but i think i fixed all your feedback :)

mortenoh:
Re-reviewed at `e56c8d6`. All five points from the previous round are addressed:

1. **"With either option" ambiguity (line 12)** -- fixed. Now reads "With both A and B ... so with Option 1 you can skip the manual URL step below", so the A/B scope and the Option 1/2 scope are both explicit.
2. **"skip the manual URL step" vs. the evaluation prerequisite** -- fixed. Line 27 is scoped with "**Option 2 only** (or if the DHIS2 Route was not registered for you)", and line 36 now says "With Option 1 this was done for you by the registered DHIS2 Route; otherwise it is the previous step". The two spots agree.
3. **Option B linked the wrong install path** -- fixed in substance. It now points at `chap-contributor-setup.md` and `make restart` instead of the pinned-release Docker install. Verified against the repo: `make restart` runs `docker compose -f compose.yml -f compose.chapkit.yml up -d --build`, `compose.yml` has `build: .` and publishes `8000:8000`, so it really does build and serve the working tree on the port docker-dhis2-core expects. See the one remaining gap below.
4. **`make start-chap` pulls the chapkit models** -- fixed. Line 9 notes it brings up the chapkit model services and that the first start pulls several extra images.
5. **Heading levels** -- fixed. Both Options are `##` now, so they sit as siblings of "After DHIS2 is up" in the sidebar.

**One thing left, non-blocking: the Option B instruction chain has a missing step.**

Line 10 sends the reader to *Setting Up Chap Core as a Contributor* and then tells them to run `make restart`. That setup page is purely a uv/venv/CLI guide -- it never mentions Docker, `compose.yml`, or a `.env` file. But `make restart` goes through `compose.yml`, which interpolates `${POSTGRES_USER}`, `${POSTGRES_PASSWORD}` and `${POSTGRES_DB}`, and `.env` is gitignored (`.gitignore:32`), so a fresh checkout does not have one. `cp .env.example .env` is documented in `README.md:28` and `docs/modeling-app/fresh-installation.md:42`, but nowhere on the path this page now sends contributors down. Following line 10 literally on a clean clone gives compose "variable is not set" warnings and a postgres container that comes up with empty credentials.

Simplest fix is to fold the step into line 10, e.g. "... set up your development environment as described in [Setting Up Chap Core as a Contributor](chap-contributor-setup.md), then in your chap-core checkout run `cp .env.example .env` (once) and `make restart` to build and serve your working tree on port 8000."

Also worth considering, though clearly out of scope for this PR: the missing `.env` step is really a hole in `chap-contributor-setup.md` itself, since that page claims to be the contributor environment setup and stops at `uv sync`. Might be worth a follow-up issue rather than patching around it here.

Nothing else changed factually -- the `/health` body, port 8000, the `host.docker.internal` default in docker-dhis2-core's `make start`, and both links all still check out.

mortenoh (approved):

mortenoh:
I'm merging this through, but please remember to ALWAYS sign your commits, this is a requirement now

PR #539, opened 2026-08-28
feat(analytics): redesign thresholds API with typed per-strategy params and multi-line responses

mortenoh:
Implements [CLIM-938](https://dhis2.atlassian.net/browse/CLIM-938). Follow-up to the Uganda percentile threshold work; supersedes the demo-only #533 as the production path for the percentile strategy.

## What changed

**Typed per-strategy params.** `POST /v1/analytics/thresholds` no longer takes `strategy: str` plus an untyped `params: dict`. Each strategy declares a pydantic params model whose `type` literal acts as the discriminator (`SeasonalParams`, `PercentileParams`), and the request's `params` field is the `ThresholdParams` discriminated union. The OpenAPI schema renders it as `oneOf` with a discriminator mapping, so the generated TS client gets a proper tagged union the modeling app can derive its strategy dropdown and params form from. `type` is required (no default), so the generated client gets a real tagged union rather than `type?: string`. Pydantic field constraints carry the validation (quantile range, positive `baselineYears`), so invalid params become 422s with field-level messages, and every field has a schema description.

The `@threshold` registry decorator now binds the params model and asserts its `type` literal matches the registered strategy id, keeping the union and the registry in sync.

**Multi-line responses.** The line parameter (`stdMultiplier`, `quantile`) accepts a scalar or a list; all entries share one baseline window and one groupby, so N lines cost roughly one computation (e.g. `quantile: [0.25, 0.75]` for the endemic channel band). The response is now an envelope: the resolved params echoed back (showing which defaults applied), `lines` with the line parameter value each threshold was computed from, and entries whose `values` array holds one nullable threshold per line, in the order of `lines`. A scalar or default request yields one-element arrays. Every requested (period, location) gets an entry; cells that cannot be computed are `null` rather than dropped, so the client never has to diff the response against the request.

```json
{
  "params": {"type": "percentile", "quantile": [0.25, 0.75], "baselineYears": 5},
  "lines": [0.25, 0.75],
  "entries": [
    {"period": "2026-01", "location": "Wakiso", "values": [1691.0, 2399.0]}
  ]
}
```

**Percentile strategy (production).** WHO-style endemic channel: a per-season percentile over the most recent `baselineYears` complete years in the dataset (default 5, `null` = all history). The channel is static: a partial final year is excluded, and the requested periods only select which rows come back, so past and future periods share one line and the result is cacheable per dataset and params. Named `baselineYears` (not lookback) per outbreak-detection literature, and `stdMultiplier` (not `k`) for the seasonal strategy, per the naming discussion on the issue.

**Shared vectorized helpers.** Season/year extraction moved to `chap_core/time_period/vectorized.py`, shared by both strategies and the outbreak-detection metrics.

## Breaking change

The request and response shapes of `POST /v1/analytics/thresholds` change. The only consumer is the modeling app's seasonal call (`useEndemicThresholds.ts`), which migrates in the same release window as the strategy dropdown / params form work. #533 remains unmerged demo code.

## Testing

- `make lint` (ruff, mypy, pyright) passes
- `make test`: 1279 passed, 118 skipped
- New tests: registry/params-literal sync, multi-line ordering for both strategies, baseline window filtering, null-baseline all-history, params validation (422), envelope shape and resolved-params echo, static line across past and future periods, partial-final-year exclusion, registry/params-union lock-in, required `type` in the schema, `lines` echo, null-filled grid for missing (period, location) combinations
- Contributor guide for custom strategies updated to the new contract (doc code blocks are mktestdocs-tested)

[CLIM-938]: https://dhis2.atlassian.net/browse/CLIM-938?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


chap_core/assessment/thresholds/params.py
```diff
@@ -0,0 +1,60 @@
...
+
+from chap_core.database.base_tables import DBModel
+
+Fraction = Annotated[float, Field(ge=0.0, le=1.0)]
+
+
+class SeasonalParams(DBModel):
+    """Parameters for the seasonal mean + k*std strategy."""
+
+    type: Literal["seasonal"] = "seasonal"
```
edvinstava: From the generated client this tagged union doesn't land. `type` has a default, so OpenAPI marks it optional and codegen emits:

```ts
export type SeasonalParams = { type?: string; stdMultiplier?: (number | Array<number>) };
```

type?: string can't discriminate SeasonalParams | PercentileParams — {} satisfies both.

Could we make type required in the schema? Same on line 39.

chap_core/assessment/thresholds/percentile.py
```diff
@@ -0,0 +1,93 @@
...
+    The baseline window is anchored on the latest requested period's year: it spans the
+    ``baseline_years`` years before that year, and all requested periods share this one
+    window. The anchor year is itself excluded, so an outbreak in progress cannot raise
+    its own threshold. When a request spans a year boundary, the earlier periods' own
+    year is still inside the window. Pass ``None`` to use all available history.
+    """
+    if baseline_years is None:
+        return historical_observations
+
+    anchor = int(extract_year(pd.Series(period_ids)).max())
```
knutdrand: The baseline window is anchored on the *request* here (`extract_year(period_ids).max()`), which makes the threshold value depend on which periods you ask for. I think that's the wrong model for this endpoint.

The intent is a **static** endemic channel: one dataset defines one threshold per `(location, season)`, and `period_ids` only selects which rows come back. That's already exactly what the `seasonal` strategy does — it computes `f(location, season)` over all history and uses `align_seasonal_to_periods` purely as a projection. `percentile` is the only strategy where the request feeds back into the value, and this line is the only place it does.

Concretely it means a multi-year request gets one window taken from the latest requested year: asking for `["2020-01" ... "2024-12"]` in one call gives the 2020 periods a 2019-2023 baseline, i.e. data from their own year and their future. And a forward-looking request — the main use case — 400s when it shouldn't: `2026-03` against a dataset ending in 2024 anchors on 2026 and finds an empty window.

Suggested change: drop `period_ids` from `filter_to_baseline` and anchor on the dataset instead.

```python
def filter_to_baseline(historical_observations, baseline_years):
    """Restrict observations to the most recent `baseline_years` complete years in the dataset."""
    if baseline_years is None:
        return historical_observations
    years = extract_year(historical_observations["time_period"])
    last = last_complete_year(historical_observations)  # excludes an in-progress final year
    windowed = historical_observations[(years > last - baseline_years) & (years <= last)]
```

`last_complete_year` is the only new piece: take `max(years)` and drop it if its season bucket count is short (< 12 months / < 52 weeks). A live dataset's final year is partial, so including it would give January's channel one more observation than November's — and excluding it preserves the property the docstring claims, that an outbreak in progress can't raise the threshold it's about to be compared against.

Follow-on effects: future periods work, past and future periods get the same line (which is what makes the retrospective view readable), the result becomes cacheable per `(dataset_id, params, locations)`, and the two strategies stop diverging — both `f(location, season)`, differing only in the statistic and whether a window is applied.

Doc updates that come with it: `PercentileParams.baseline_years` ("Number of complete years before the requested periods..." -> "the most recent N complete years in the dataset"), and the year-boundary caveat in this docstring goes away entirely.

One consequence worth stating in the module docstring rather than leaving to be re-derived as a bug later: a static channel scores past periods against a baseline that includes them, so a historical epidemic partly raises the line it is plotted against. That is inherent to the endemic-channel idiom and fine given the past view is visual.

chap_core/assessment/thresholds/params.py
```diff
@@ -0,0 +1,60 @@
...
+    )
+    baseline_years: int | None = Field(
+        5,
+        ge=1,
+        description="Number of complete years before the requested periods to compute the baseline from. "
+        "`null` uses all available history.",
+    )
+
+
+ThresholdParams = Annotated[SeasonalParams | PercentileParams, Field(discriminator="type")]
```
knutdrand, line 75: Nothing couples this union to the registry, so the two can drift silently. `threshold()` asserts `params_model.type == strategy_id`, but not that `params_model` is actually a member of `ThresholdParams`. A strategy that is registered and imported in `_discover_strategies` but not added here is advertised by `GET /thresholds/strategies` and then rejected with a 422 when a client builds the request the picker just told it to build — the strategy list and the request schema disagree, and neither the strategy author nor CI notices.

It also undercuts the claim in the module docstring that "the endpoint code never needs editing": adding a strategy needs a manual edit in two places, and this is the one with no failure mode attached to forgetting it.

Smallest fix is to assert it next to the literal check already in the decorator, so it fails at import with a clear message rather than at request time with a 422:

```python
from chap_core.assessment.thresholds.params import ThresholdParams  # no cycle: params imports nothing from here

union, _ = get_args(ThresholdParams)
if params_model not in get_args(union):
    raise ValueError(
        f"{params_model.__name__} is not a member of ThresholdParams; "
        f"add it to the union in thresholds/params.py"
    )
```

The alternative is to build the union from the registry in a module imported after `_discover_strategies()`, which removes the second edit entirely — the OpenAPI schema is still correct since pydantic builds it at runtime, but static checkers lose the union, which seems a bad trade given typed codegen is the point of this design.

There is also precedent for the lock-in route: `@backtest_plot` has the same registry-plus-generated-artifact shape and guards it with `tests/test_generated_plot_ids.py`. A test asserting `set(registry) == {m.model_fields["type"].default for m in union members}` would do the same job here if you would rather keep the decorator untouched.

chap_core/rest_api/v1/routers/analytics.py
```diff
@@ -721,21 +723,37 @@ class ThresholdRequest(DBModel):
...
+        description="One threshold per requested line, in the order of the request's line parameter list "
+        "(`quantile`, `stdMultiplier`, ...). A scalar or default request yields one element. "
+        "An element is `null` when that line could not be computed."
+    )
+
 
-    period: str = Field(description="Period the threshold applies to.")
-    location: str = Field(description="Location the threshold applies to.")
-    value: float | None = Field(description="Computed threshold value, or `None` if it could not be computed.")
+class ThresholdResponse(DBModel):
```
edvinstava, line 807: could we add `lines: list[float]` with the numbers used to compute each threshold (the quantiles or std multipliers), in the same order as `values`?

Then the frontend can match each threshold to its number without knowing the strategy, or whether the request sent one number or several.

chap_core/assessment/thresholds/base.py
```diff
@@ -12,36 +13,63 @@
...
+    import pandas as pd
+
+    from chap_core.time_period.vectorized import season_column
+
+    requested = pd.DataFrame({"period_id": period_ids})
+    season, buckets = season_column(requested["period_id"])
+    if season not in per_season.columns:
+        raise ValueError(f"period_ids are {season}-based but the dataset's time periods have a different frequency")
+    requested[season] = buckets
+    merged = requested.merge(per_season, on=season)
```
edvinstava, line 75: I think this makes a requested period with no matching baseline newer becomes a row. Dropping the row instead of returning null leads to the fronend having to diff the ressponse and request to see if antyhing is missing. 

Could we instead have it return the full requested and leave missing cells at null?

edvinstava (commented):
Overall looks very good and works well! Juat have a few comments that can make it a bit simpler and improve the frontend side of this change. Let me know when you have fixed it up so I can update the frontend again 🙌🏼

ivargr (approved):

PR #549, opened 2026-09-02
fix: restrict all locations dataset when last_n is set for explain-lime

willbehn:
Fixed a bug in producing datasets for LIME explanations when the last_n argument was set.

Only the explained locations dataset was restricted when last_n was set. The issue is that this dataset was then spliced back together with the other locations datasets, which no longer matched in dimensionality because it had been restricted.

### DIFF
--- a/chap_core/cli_endpoints/explain.py
+++ b/chap_core/cli_endpoints/explain.py
@@ -6,7 +6,7 @@ import logging
 from typing import Annotated
 
 from cyclopts import Parameter
-from pydantic import BaseModel
+from pydantic import BaseModel, Field
 
 from chap_core.api_types import RunConfig
 from chap_core.cli_endpoints._args import (  # noqa: TC001 — used at runtime via cyclopts get_type_hints()
@@ -36,7 +36,16 @@ class LimeParams(BaseModel):
     seed: int | None = None
     timed: bool = False
     adaptive: bool = False
-    last_n: int | None = None
+    last_n: int | None = Field(
+        default=None,
+        description=(
+            "If set, restricts every location's historical data fed to the model to the last "
+            "last_n time steps of the explained location's own timeline (other locations may end "
+            "up with fewer steps, or be dropped, if their range doesn't reach that far back). Does "
+            "not affect the background sampler's pool or the per-feature global means, which still "
+            "draw from the full unrestricted dataset."
+        ),
+    )
     with_metrics: bool = False
 
 
--- a/chap_core/cli_endpoints/explain.py
+++ b/chap_core/cli_endpoints/explain.py
@@ -39,7 +39,7 @@ class LimeParams(BaseModel):
     last_n: int | None = Field(
         default=None,
         description=(
-            "If set, restricts every location's historical data fed to the model to the last "
+            "If set, restricts every location's historical data fed to the model to the "
             "last_n time steps of the explained location's own timeline (other locations may end "
             "up with fewer steps, or be dropped, if their range doesn't reach that far back). Does "
             "not affect the background sampler's pool or the per-feature global means, which still "
--- a/chap_core/explainability/lime.py
+++ b/chap_core/explainability/lime.py
@@ -977,8 +977,8 @@ def _prepare_lime_inputs(
     climate_predictor = get_climate_predictor(climate_data)
     full_future_weather = climate_predictor.predict(prediction_range)
 
-    # Restrict every location's history to each locations own last last_n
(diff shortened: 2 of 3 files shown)


willbehn:
@mortenoh @knutdrand

mortenoh:
## Code review

The core fix is right. Restricting only the target location's `hist_df` and then splicing it back into an unrestricted `full_dataset` did produce mismatched series lengths; truncating the whole `DataSet` up front fixes that consistently, and the baseline `y_orig` runs use the same truncated dataset so eLoss stays comparable.

Five findings, roughly in severity order:

**1. `chap_core/explainability/lime.py:984` - window derived from the wrong location.**
`window`/`start_period` come from `dataset.period_range`, which returns only the *first* location's `time_period` (`temporal_dataclass.py:281`), not the target's or a union. `DataSet` explicitly allows per-location ranges to differ (`start_timestamp`/`end_timestamp` are `min`/`max` over locations). If district A runs through 2024-12 and district B only through 2023-06, `last_n=12` gives `start_period` 2024-01, and restricting B clamps `searchsorted` to `len` and yields an empty series - which then goes straight into `model.predict` at lines 727-730. The same cause means the explained window is not exactly `last_n` steps whenever the target's range differs from the first location's. Suggest deriving the window from `dataset[location].time_period` and skipping locations that come back empty.

**2. `chap_core/explainability/lime.py:1031` - background sampler and `global_means` silently change behaviour.**
`full_dataset_df = dataset.to_pandas()` now sees the truncated dataset. In `RandomBackground.sample` (`perturb.py:200-209`) every location's frame is exactly `last_n` rows and segment lengths are drawn from an equally long `hist_df`, so `max_start = len(loc_df) - length` collapses toward 0 - the random background window becomes the same aligned window on every draw, removing the sampling diversity the perturbation scheme depends on. Additionally `float(full_dataset_df[feat].mean())` can now be `NaN` for a feature entirely missing within the last `last_n` periods (realistic with reporting lag), and that propagates into the static-feature fallback and hence into `x0` and the surrogate fit. If truncating the background pool was not intended, keep an unrestricted copy of the dataset for `full_dataset_df`/`global_means`.

**3. `chap_core/explainability/lime.py:986` - semantic change worth an explicit decision.**
The truncated dataset is what `produce_lime_dataset` hands to `model.predict` for every perturbation, so the black-box model is now conditioned on only `last_n` periods of history for all locations, not just the explained one. With a typical `last_n` (e.g. 12 months) most CHAP models lose the seasonality they need, and the resulting explanation describes a differently-conditioned model than the one that was trained. If the goal is "explain the recent window" rather than "condition on the recent window", restrict only the segmentation/`x0` window and keep the target location at full length.

**4. `chap_core/explainability/lime.py:1090` and `:1375` - stale docstrings.**
Both `last_n` docstrings still say only the last `last_n` steps of *the location's* historical data are used. That is now wrong: every location is truncated, and the truncation feeds the model, the sampler and the global means. Same wording applies to the CLI option backing `LimeParams.last_n` in `chap_core/cli_endpoints/explain.py:39`.

**5. No test accompanies the fix.**
`grep -rn "last_n"` hits only `lime.py` and `cli_endpoints/explain.py`, so the exact regression being fixed has no coverage and can silently return. Per CLAUDE.md rule 9, a small test asserting every location in the prepared dataset has the same length when `last_n` is set would lock this in.

Findings 1-3 are behavioural and worth resolving before merge; 2 and 3 in particular may want a call on intent first.

willbehn:
Thank you for your thorough review @mortenoh !
I have addressed the findings and made the following changes:
Issue 1 has been addressed, the window was being derived from the wrong location, and this has now been fixed. Locations are now also dropped if their time period after last_n doesn't match the target's window length (not just when they come back empty).
Issues 4 and 5 have also been addressed.
Regarding issues 2 and 3
* I have made the change so that the sampler now sees the original dataset, before it has been truncated. I believe this is a serviceable fix for issue 2 at this point.
* I have not addressed issue 3 in this PR. I believe this is a larger fix than what this PR was intended to cover. This PR's intention was to make the explain-lime CLI run with the last_n parameter without crashing, not to redefine what last_n means for the model's predictions.

However, I am working on explainability in CHAP as part of my master's degree, and will most likely come back to this later. If this is something you want fixed before the changes are merged, I can gladly look into it as well and am happy to discuss it further. But I believe it is outside the original scope I intended for this PR.

mortenoh (approved):

willbehn:
@mortenoh I rebased after your approval to add commit signatures (required for merge), no code is changed, might need to re-approve. 
Thank you!

PR #550, opened 2026-09-02
feat(compose): make compose.ghcr.yml a self-contained deployment file

mortenoh:
[CLIM-975](https://dhis2.atlassian.net/browse/CLIM-975)

## Problem

`compose.ghcr.yml` is meant to be the deployment path, but two things stopped it being one:

- The images were pinned to `chap-core:latest` / `chap-worker:latest`, so a specific release could not be deployed without editing the file.
- `POSTGRES_USER`, `POSTGRES_PASSWORD` and `POSTGRES_DB` had no defaults, so it only worked next to an `.env` from a source checkout, which is the thing a prebuilt-image deployment is supposed to avoid.

It had also drifted from `compose.yml` on three points that only affect the deployment path.

## Change

Deploying a release is now one file and one command, with nothing else to configure:

```console
curl -O https://raw.githubusercontent.com/dhis2-chap/chap-core/master/compose.ghcr.yml
docker compose -f compose.ghcr.yml up -d
```

- Image tag is `${CHAP_IMAGE_TAG:-latest}` on both services.
- Every interpolation has a default, so the file renders and runs as downloaded.
- The header and the docs list all seven supported variables with their defaults, so it is clear both that nothing is required and what can be tuned: `CHAP_IMAGE_TAG`, `POSTGRES_USER`, `POSTGRES_PASSWORD`, `POSTGRES_DB`, `CHAP_DATABASE_URL`, `CHAP_ROOT_PATH`, `CHAP_API_TOKEN`.
- `CHAP_DATABASE_URL` overrides the composed URL, for a password that is not URL-safe.
- Drift closed against `compose.yml`: `init: true` on both services, `CHAP_ROOT_PATH` on `chap`, and healthchecks declared in the file rather than inherited from the images, since a pinned tag can predate them.
- Docs get a "Deploying a release without a checkout" section with a settings table.

`compose.yml` is untouched: it stays the unambiguous build-from-source file for local development, and the two remain alternatives rather than stackable.

### Included fix: `.env.example` shipped a Redis password

Separate cause, same file set. `.env.example` set `REDIS_PASSWORD=your_secure_redis_password`, but the compose valkey service runs without `--requirepass`. `chap_core/util.py:101` reads the variable and sends AUTH to a server with no password, which valkey rejects — so the documented "copy `.env.example` to `.env`" flow was broken. CI had been deleting the line after copying (`ci-test-external-models.yml`); the cause is fixed and that workaround removed.

Leaving the compose valkey unauthenticated is tracked separately as [CLIM-977](https://dhis2.atlassian.net/browse/CLIM-977): it is not reachable from the host, but every service on the default bridge, model containers included, can reach the broker.

The Kubernetes chart is the opposite case: its valkey does require a password, and `skaffold run` reads `REDIS_PASSWORD` from the shell (`skaffold.yaml:26`). Since `.envrc` is `dotenv`, leaving the value in `.env.example` fed it to both. The comment now points the k8s flow at the shell, as `charts/chap/README.md` already does. No chart file changes.

## Review follow-ups

- **Tag persistence** — the docs recommended `CHAP_IMAGE_TAG=v1.2.3 docker compose ...`, which applies to one command, so a later `pull` or `up` fell back to `latest`. They now direct the pin to an `.env` file and say why.
- **Password in the URI** — confirmed at `chap_core/database/database.py:52`, which passes the value straight to `create_engine`. Hence the `CHAP_DATABASE_URL` override, and the URL-safe note.
- **Worker healthcheck on old tags** — `chap-worker:v2.1.0` has no HEALTHCHECK (added later, in #532), so a pinned old tag ran unmonitored. Now declared in the compose file. Note the review's premise that v2.1.0's own compose file supplied one does not hold — `git show v2.1.0:compose.ghcr.yml` has no healthcheck on the worker — so this was a gap rather than a regression.

## Tests

Four in `tests/test_compose_deployment.py`: the standalone file has no undefaulted variable, `chap` and `worker` each carry a pinnable tag and their own compose-level healthcheck, and `.env.example` leaves `REDIS_PASSWORD` unset while the compose valkey takes no password. The image-to-Dockerfile healthcheck lookup was taught to strip a templated tag.

## Verification

- `make lint` clean (ruff, mypy 416 files, pyright 0 errors)
- `make test`: 1340 passed, 120 skipped, 4 xfailed, 1 xpassed
- `docker compose -f compose.ghcr.yml config -q` exits 0
- Copied the file alone into an empty directory with no `.env`: renders and validates; `CHAP_IMAGE_TAG=v1.2.3` resolves both images to `:v1.2.3`; `CHAP_DATABASE_URL` override renders a percent-encoded password unchanged

[CLIM-975]: https://dhis2.atlassian.net/browse/CLIM-975?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CLIM-977]: https://dhis2.atlassian.net/browse/CLIM-977?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -11,3 +11,7 @@ REDIS_PASSWORD=your_secure_redis_password
 # CHAP_API_TOKEN=
 # Shared secret for chapkit service registration (optional, separate from CHAP_API_TOKEN).
 # SERVICEKIT_REGISTRATION_KEY=
+
+# Docker image tag used by compose.ghcr.yml. Defaults to `latest`; set it to a
+# release tag (e.g. v1.2.3) to deploy a specific version.
+# CHAP_IMAGE_TAG=
--- a/.env.example
+++ b/.env.example
@@ -3,8 +3,10 @@ POSTGRES_USER=chap
 POSTGRES_PASSWORD=chap
 POSTGRES_DB=chap_core
 
-# Valkey/Redis
-REDIS_PASSWORD=your_secure_redis_password
+# Valkey/Redis. Leave unset for compose and host tests: their valkey has no
+# password and rejects AUTH. The k8s chart does use one, and `skaffold run` reads
+# it from the shell instead (see charts/chap/README.md).
+# REDIS_PASSWORD=
 
 # API authentication (optional). Unset means no authentication.
 # Generate with `openssl rand -hex 32`; use at least 32 characters.
@@ -12,11 +14,8 @@ REDIS_PASSWORD=your_secure_redis_password
 # Shared secret for chapkit service registration (optional, separate from CHAP_API_TOKEN).
 # SERVICEKIT_REGISTRATION_KEY=
 
-# Docker image tag used by compose.ghcr.yml. Defaults to `latest`; set it to a
-# release tag (e.g. v1.2.3) to deploy a specific version. Set it here rather
-# than inline, so every later compose command keeps the same pin.
+# Image tag for compose.ghcr.yml. Set a release tag (e.g. v1.2.3) to pin a version.
 # CHAP_IMAGE_TAG=
 
-# Full database URL, overriding the one composed from the POSTGRES_* values.
-# Needed when POSTGRES_PASSWORD is not URL-safe; percent-encode it here.
+# Overrides the URL composed from POSTGRES_*. Needed for a non-URL-safe password.
 # CHAP_DATABASE_URL=
--- a/.env.example
+++ b/.env.example
@@ -13,5 +13,10 @@ REDIS_PASSWORD=your_secure_redis_password
 # SERVICEKIT_REGISTRATION_KEY=
 
 # Docker image tag used by compose.ghcr.yml. Defaults to `latest`; set it to a
-# release tag (e.g. v1.2.3) to deploy a specific version.
+# release tag (e.g. v1.2.3) to deploy a specific version. Set it here rather
+# than inline, so every later compose command keeps the same pin.
 # CHAP_IMAGE_TAG=
+
+# Full database URL, overriding the one composed from the POSTGRES_* values.
+# Needed when POSTGRES_PASSWORD is not URL-safe; percent-encode it here.
+# CHAP_DATABASE_URL=
--- a/.github/workflows/ci-test-external-models.yml
+++ b/.github/workflows/ci-test-external-models.yml
@@ -84,9 +84,4 @@ jobs:
       - name: Run
         run: |
           cp .env.example .env
-          # The placeholder REDIS_PASSWORD from .env.example gets loaded via
-          # load_dotenv() at import time and makes the redis client send AUTH
-          # to the unauthenticated CI valkey service. Strip it so the host
-          # tests match the no-password setup.
-          sed -i '/^REDIS_PASSWORD=/d' .env
           make test-all
--- a/compose.ghcr.yml
+++ b/compose.ghcr.yml
@@ -1,11 +1,17 @@
 # Deployment file. Download this file on its own and run `docker compose up -d`:
 # it pulls prebuilt images from GHCR and needs no source checkout and no .env.
 #
-# Pin a release with CHAP_IMAGE_TAG, which defaults to `latest`:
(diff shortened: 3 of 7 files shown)


mortenoh:
All three addressed in 3390d1e3.

**[P1] CHAP_IMAGE_TAG persistence** — correct, and the docs were actively encouraging the broken form. The inline assignment applies to one invocation, so a later `pull` or `up` resolves back to `latest` and replaces the pinned release. The docs and `.env.example` now tell you to put the pin in an `.env` file beside the compose file, which Compose reads on every command, and say explicitly why the inline form is not enough.

**[P2] Password in the database URI** — confirmed: `chap_core/database/database.py:52` hands the value straight to `create_engine`, so `strong@password` splits at the wrong `@`. `CHAP_DATABASE_URL` now overrides the composed URL (`${CHAP_DATABASE_URL:-postgresql://...}`), so a password with reserved characters is set once, percent-encoded, in one place. The header comment and docs name the unsafe characters and show the encoded form. Verified both paths render:

```
default:   postgresql://chap:chap@postgres:5432/chap_core
override:  postgresql://chap:p%40ss@postgres:5432/chap_core
```

**[P3] Worker healthcheck on pinned older images** — the conclusion holds, though one premise does not. `Dockerfile.worker` gained its HEALTHCHECK in #532, which is not an ancestor of v2.1.0, so `chap-worker:v2.1.0` indeed has none. But that release's `compose.ghcr.yml` did not supply one either (`git show v2.1.0:compose.ghcr.yml` — the worker service has no healthcheck block), so this is not a regression against it. It is still a real gap for a pinned old tag, so the worker healthcheck is now declared in `compose.ghcr.yml` rather than inherited from the image. `/app/.venv/bin/celery` exists back to v2.1.0, so the check works on the older images too.

A test locks the last one in: `test_standalone_compose_declares_own_healthcheck` requires both `chap` and `worker` to carry a compose-level healthcheck in the standalone file, so the image can never quietly become the only source again.

`make lint` clean; `make test` 1339 passed, 120 skipped, 4 xfailed, 1 xpassed.

mortenoh:
Ran the actual deployment, not just `config`. No merge needed for this: the file pulls images that are already published, so the deploy path is exercisable from the branch.

**Default, exactly as a user would do it** — copied only `compose.ghcr.yml` into an empty directory, no `.env`, no checkout:

```
SERVICE    STATUS
chap       Up (healthy)
postgres   Up (healthy)
redis      Up (healthy)
worker     Up (healthy)
```

- `GET /health` -> 200 `{"status":"success","message":"healthy"}`, `GET /docs` -> 200
- Alembic ran the migrations against the composed database URL, so the `POSTGRES_*` defaults wire up correctly
- Worker: `Connected to redis://redis:6379//`, `celery@... ready`
- Images resolved to `ghcr.io/dhis2-chap/chap-core:latest` and `chap-worker:latest`

**Pinned release** — same file, `.env` containing only `CHAP_IMAGE_TAG=v2.1.0`:

```
SERVICE    IMAGE                                   STATUS
chap       ghcr.io/dhis2-chap/chap-core:v2.1.0     Up (healthy)
worker     ghcr.io/dhis2-chap/chap-worker:v2.1.0   Up (healthy)
```

`/health` returns 200 on the pinned version too, and this is the direct check on the P3 fix:

```console
$ docker inspect ghcr.io/dhis2-chap/chap-worker:v2.1.0 --format '{{if .Config.Healthcheck}}...{{end}}'
image declares NONE (compose supplied it)
```

The v2.1.0 worker image carries no HEALTHCHECK of its own, and it reports healthy purely from the healthcheck this file now declares. That confirms `/app/.venv/bin/celery ... inspect ping` works on the older image as well.

Both stacks torn down with `down -v`.

PR #552, opened 2026-09-04
added optimization_direction in MetricSpec, extra argument search_spa…

zlilu:
Added optimization_direction in MetricSpec, extra argument search_space_yaml in eval_cmd, max_trials and seed through estimator-options.


knutdrand:
## Review

Reviewed `edc4f26d..131b08f1`. `make lint`, `mypy` and the metric/eval/hpo tests pass on the branch.

### Request: move `--hpo-search-space-yaml` under `--estimator-options`

`chap eval` already has a long top-level arg list, and this PR adds another path arg to it. `--hpo-search-space-yaml` is only meaningful in HPO mode, so it belongs next to the other HPO knobs as `--estimator-options.hpo-search-space-yaml` rather than at the top level.

`--model-configuration-yaml` stays where it is - it applies in all modes and is shared with `explain`, `report` and `causal`.

### Findings

**1. High - `chap_core/hpo/hpoModel.py:66` - `self._leaderboard` is no longer reset in `get_leaderboard()`**

The old code cleared the list before the search loop; the rewrite dropped it. `get_leaderboard()` is public API (abstract on `HpoModelInterface`), so a second call appends onto the stale list. The returned leaderboard then has duplicate/stale rows, and the `assert best_params == self._leaderboard[0]["config"]` on line 108 can fire *after* every expensive trial has already run:

```python
m = HpoModel(objective=..., searcher=GridSearcher(), configuration=None,
             search_space={"x": [1, 2]}, max_trials=None, seed=None)
m.get_leaderboard("ds")   # x=1 -> 1.0, x=2 -> 9.0
m.get_leaderboard("ds")   # x=1 -> 5.0, x=2 -> 2.0
# AssertionError: best params is not the first in leaderboard
```

**2. High - `chap_core/hpo/hpoModel.py:29,48` - `configuration` is stored but never used, so `--model-configuration-yaml` is silently dropped in HPO mode**

`train()` validates `self._best_config`, which only ever contains `{"user_option_values": <searched params>}`; the merge with `self._configuration` is commented out at lines 51-55. Concretely: `chap eval --estimator-options.mode hpo --model-configuration-yaml cfg.yaml` with `cfg.yaml` setting `additional_continuous_covariates: [population]` - `warn_unused_covariates` sees the covariate and stays quiet, but `get_train_predict_runner_from_model_template_config` writes `model_configuration_for_run.yaml` from the HPO config with `additional_continuous_covariates: []`. The model trains without the covariate and the user gets no warning. Any fixed `user_option_values` in that file are dropped the same way.

**3. Medium - `chap_core/cli_endpoints/evaluate.py:274` - HPO runs record the pre-tuning configuration**

`configured_model_db` is built from the user's base `configuration` rather than from the estimator's best configuration, so the exported `.nc`/backtest metadata for an HPO run reports `user_option_values: {}` instead of the hyperparameters actually selected. Same root cause as #2.

**4. Medium - `chap_core/hpo/searcher.py:104` + `hpoModel.py:76` - `RandomSearcher` + `max_trials=None` never terminates**

`RandomSearcher.ask()` is now typed and implemented as always returning a `SearchCandidate` (the `emitted >= max_trials` stop moved into `HpoModel`), while `HpoModel.__init__` accepts `max_trials: int | None`. `HpoModel(..., searcher=RandomSearcher(), max_trials=None)` loops forever. `get_hpo_estimator` always supplies a default for random/TPE today, so this is only reachable via direct construction - but `HpoModel` is a public keyword-arg constructor and nothing guards the combination. Worth rejecting `max_trials=None` for non-exhausting searchers.

**5. Medium - `chap_core/assessment/metrics/base.py:52` - `optimization_direction` is a required `MetricSpec` field with no default**

It's inserted as the third *positional* field, ahead of the defaulted `output_dimensions`/`aggregation_op`/`description`. Any out-of-tree metric (which `docs/contributor/creating_custom_metrics.md` actively encourages) that constructs `MetricSpec(...)` without the new kwarg now raises `TypeError` at import, and positional construction `MetricSpec("id", "name", (DataDimension.location,))` silently binds the dimensions tuple to `optimization_direction`. Giving it `= None` and moving it after the other defaulted fields is behaviourally identical for all in-repo call sites and keeps the change additive.

**6. Medium - deleted HPO tests with no replacement**

`tests/hpo/test_hpo_estimator.py`, `test_hpo_model.py` and `test_objective.py` are removed ("deleted hpo tests for later refactor"), while the PR simultaneously introduces new behaviour: `max_trials`, `seed`, `OptimizationDirection`-driven comparison, `SearchCandidate`, and the new `get_hpo_estimator` keyword API. That leaves `tests/hpo/` covering only base utils plus one TPE ask/tell. Findings #1 and #4 would both have been caught by the deleted `test_hpo_model.py`. Per CLAUDE.md rule 9 the changed behaviour needs tests before merge.

**7. Low - `chap_core/cli_endpoints/evaluate.py:175` - docstring example uses a flag the CLI doesn't accept**

`--hpo_search_space_yaml` (underscores) isn't registered; cyclopts only accepts `--hpo-search-space-yaml`. Copy-pasting the documented example fails. The `--estimator_options.metric` / `.searcher` lines just below have the same problem (pre-existing, but easy to fix in the same pass).

**8. Low - `tests/hpo/test_searcher.py:12` - assertion checks the wrong candidate**

`assert p2 is not None and p1.token is not None` re-checks `p1`; should be `p2.token`. The second `ask()`'s token is never verified, which is exactly what the `_trial_id` -> `SearchCandidate.token` refactor changed.

### Checked, not a problem

- `**configuration.model_dump() if configuration else {}` parses as `**(... if ... else {})` and both keys are real `ConfiguredModelDB` fields - this is a fix over the previous `configuration=<dict>`, which `extra="forbid"` only tolerated because SQLModel `table=True` skips validation.
- `SearcherType` is still imported at runtime in `_common.py:295` despite being dropped from the `TYPE_CHECKING` block.
- All 20 in-repo `MetricSpec(...)` sites updated; registry directions verified at runtime.
- No remaining callers of `get_best_config`, `DEFAULT_SEARCH_TRIALS`, or the old positional `HpoModel(...)` / `Searcher.ask()` contract.

knutdrand (approved):
Approving. Most of the review is addressed in fd175c71: leaderboard reset restored, `max_trials` guard for non-exhaustive searchers, `MetricSpec.optimization_direction` defaulted and moved last, `test_searcher.py` assertion corrected, and the search space moved onto `--estimator-options` as asked.

I pushed one commit (8c2d02d1) replacing the `model._max_trials` read in `test_hpo_model.py` with a `get_leaderboard()` assertion, per CLAUDE.md rules 9 and 13.

Three items from the review are still open - fine as follow-ups, but they should not get lost:

1. **HPO metadata still records the pre-tuning configuration.** The merge in `train()` mutates `self._configuration` in place, but `evaluate.py:272` has already called `configuration.model_dump()` to build `configured_model_db` before `Evaluation.create` (line 293) triggers `train()`. The exported `.nc` still carries the untuned `user_option_values`.

2. **The documented flag name does not exist.** The field is `EstimatorOptions.search_space`, so cyclopts registers `--estimator-options.search-space`. Three strings name something else: the help text at `evaluate.py:133` (`search_space_yaml`), the docstring example at `evaluate.py:172` (`search-space-yaml`), and the error message at `_common.py:311` (`search-space-yaml`).

3. **Test coverage.** `test_hpo_estimator.py` and `test_objective.py` are still deleted relative to master, and nothing covers the leaderboard reset or the configuration merge.

PR #553, opened 2026-09-04
docs(lime): clean up last_n docs, fix surrogate_name options

willbehn:
Small documentation cleanup following #549, and removal of stale option:
* Fixed the default value for last_n in the lime CLI.
* Removed "tree" from surrogate_name's listed options, it was not implemented.
* Rewrote the last_n docstrings in explain()/explain_adaptive() to be clearer
* Fixed a few typos in comments and test docstrings.


willbehn:
@mortenoh

mortenoh:
Thanks for the cleanup. Since this is docs-only nothing here can break at runtime; the findings below are documentation regressions and gaps against the PR's own stated goals. I would ask for 1 and 2 before merge, 3 is a follow-up in the same spirit, 4 and 5 are nits.

**1. `chap_core/cli_endpoints/explain.py` line 39 -- dropping the `Field(description=...)` regresses `--help`**

Removing the pydantic `Field` description means the rendered `chap explain-lime --help` no longer has a `LAST-N` row of its own. On master that row reads "If set, restricts every location's historical data fed to the model to the last_n time steps of the explained location's own timeline (other locations may ... be dropped ...)". On this branch it is replaced by the shared "LIME pipeline configuration. Use --lime-params.granularity ..." boilerplate, and the only last_n-specific wording left is "to enable truncation of the dataset". A CLI user can no longer learn from `--help` that other locations get truncated or dropped, which is exactly the surprising behaviour the removed text warned about. Keeping `Field(default=None, description=...)` preserves the row.

**2. `chap_core/explainability/documentation.md` line 27 -- still lists `"tree"`**

The PR description says `tree` is removed from the documented `surrogate_name` options, but this file still says `one of ["ridge", "tree"]`. `disambiguate_surrogate` (`chap_core/explainability/lime.py` around line 746) only accepts `ridge`, `bayesian`, `blr` and `bayesian_linear`, so a reader following this file gets `ValueError('Unknown surrogate model: tree')` after the dataset has loaded. The file is hand-maintained, so it needs the same `["ridge", "bayesian"]` fix as the two docstrings in `lime.py`.

**3. `chap_core/explainability/userguide.md` line 116 and `docs/chap-cli/explain-lime-reference.md` line 44 -- user-facing `last_n` docs are still incomplete**

Both still describe `last_n` as restricting only the explained location's historical data. The new `lime.py` docstring correctly states that every other location is truncated to the same window and dropped if it does not cover it. The mkdocs-published reference (`| --lime-params.last-n | If set, only use the last N time steps of historical data |`) says nothing about dropping, so a user with a staggered multi-location dataset will see the "Dropping location(s) that don't fully cover ..." warning and a model fit on fewer locations with no documentation explaining why. Worth folding into this PR given its intent.

**4. `chap_core/cli_endpoints/explain.py` line 92 -- `--lime-params.last_n` vs `--lime-params.last-n`**

The new help text spells the flag with an underscore while cyclopts renders it (and every neighbouring flag in the same sentence: `--lime-params.segmenter-name`, `--lime-params.num-perturbations`, ...) with a dash. Both spellings are accepted on the command line, so this is cosmetic, but it disagrees with `docs/chap-cli/explain-lime-reference.md` which documents `--lime-params.last-n`.

**5. `chap_core/explainability/lime.py` line 1105 -- surrogate aliases**

The docstring says `surrogate_name` is one of `["ridge", "bayesian"]`, but `disambiguate_surrogate` also accepts `blr` and `bayesian_linear`. Either list them or point at `disambiguate_surrogate`.

Verified along the way: the new docstring's claim that only the model input is restricted (not the sampler pool or global means) matches `_prepare_lime_inputs`, the `Field` import removal is clean, and `make lint` passes on the branch.

willbehn:
@mortenoh Thank you for the feedback!

I have addressed all comments, except the first regarding the pydantic Field. I added it in my last pull request with the last_n fix, but I reverted it since none of the other LimeParams have their own Field. For that reason, and because the last_n behaviour is well documented elsewhere now, I think it's fine to remove for clarity.

mortenoh:
Second pass. Points 2 to 5 from the first round are addressed, thanks. On point 1, I would push back on the reasoning rather than the outcome: the convention argument goes the other way, since `chap_core/api_types.py` gives its request models per-field `Field(description=...)`, and `LimeParams` was the odd one out for having none. The consistent fix is descriptions on every `LimeParams` field so `--help` gets one row per flag. I am fine leaving that for a follow-up, but as the diff stands the `--help` output for `--lime-params.last-n` regressed from a specific sentence to the generic blob, so this PR removes CLI documentation while adding it elsewhere.

Two things I would fix before merge, since they are in the wording this PR introduces:

**1. `chap_core/explainability/lime.py` line 1111 (and 1398, `userguide.md` line 116, `explain-lime-reference.md` line 44) -- the drop rule is misdescribed**

"Locations that don't fit into the last_n period will be dropped" reads as "dropped if their history does not reach far enough back". The code slices with `slice(start_period, None)` and then drops on `len != window`, so a location that fully covers the window but extends past the explained location's last period is dropped as well. Reproduced: alpha (target) 2020-01 to 2021-08, beta 2020-01 to 2021-12, `last_n=6` gives restricted lengths alpha 6, beta 10, and beta is dropped with a warning saying it does not fully cover alpha's window, although it does. The docs should say the rule is exact alignment with the explained location's window, or the code should truncate to the target's last period instead of dropping. Related and undocumented at every site: the window is `min(last_n, len(target history))`, so `last_n` larger than the target's history is not a no-op; it still truncates and drops other locations to the target's full length.

**2. `chap_core/cli_endpoints/explain.py` line 92 -- the new help clause**

"drops locations that dont fit into last_n" has the typo and is a comma-spliced fragment, and because it lives in the shared `Parameter(help=...)` blob it renders under all eleven `--lime-params.*` flags, including `seed` and `timed`. It also says "truncation of the dataset", which overstates it: the sampler pool and global means still use the full dataset, as the docstring itself says. Suggest: "--lime-params.last-n to restrict model input to the explained location's last N steps (other locations are aligned to that window or dropped)".

Smaller, take or leave:

- `lime.py` line 1110: "not background sampler's pool etc." lost the specific "or the per-feature global means" from the old text, which is still true and is the part a reader is most likely to get wrong.
- `lime.py` line 1108 and 1395: in the same Args block you fixed for `surrogate_name`, `sampler_name` still says `one of ["background"]` while `disambiguate_sampler` accepts background, linear, constant, local_mean, global_mean, random and fourier. The userguide example uses fourier.
- `userguide.md` line 116: "based on the selected location's last_n" reads as if last_n were per location. It is one integer; the window is anchored to the selected location's timeline. With `--all-locations` the CLI calls `explain()` per location, so the cutoff and dropped set are recomputed per call, which is worth one sentence.
- PR description says "Fixed the default value for last_n"; master already had `default=None`, so the only effect of that hunk is the help text change.

Follow-ups, not this PR:

- `lime.py` line 1004: the new comment says the per-location datasets are "spliced together before being passed to model.predict()", but `full_future_weather` is returned unfiltered after the drop, so `produce_lime_dataset` still builds future rows for locations that no longer have history. Pre-existing and untested (the last_n test only records `historic_data.locations()`), but the comment now promises something the code does not do. Worth its own ticket.
- `chap_core/explainability/documentation.md` is headed "TODO: Very outdated", is not in the mkdocs nav, and still documents a `threshold` argument that does not exist and a wrong granularity default next to the line you fixed. I would delete it in favour of `userguide.md` and the CLI reference rather than keep patching single lines.

mortenoh (approved):

PR #554, opened 2026-09-06
moved train-test-generator out of backtest to Evaluation.create

zlilu:


knutdrand (approved):
Reviewed the refactor. Hoisting `train_test_generator` out of `backtest` and up into the three call sites is a clean separation, and deriving `last_train_period` from the already-materialised `train_set` rather than regenerating the splits is a genuine improvement. Split-index math, `train_set` metadata and the retrain schedule are unchanged.

Merged master in and resolved the conflict in `run_backtest`: kept `n_retrain` on the `backtest` call (new in #569) and dropped `stride`/`weather_provider`, which now go to `train_test_generator`. Full suite and lint green locally.

Two pre-existing issues this refactor puts within easy reach, worth a follow-up rather than blocking here:

- `cli_endpoints/evaluate.py` `--dry-run` does not forward `n_retrain` to `backtest`, so a dry run always trains once and never exercises the retrain-on-`historic_data` path the real run will take.
- `rest_api/db_worker_functions.py` sets `last_train_period = dataset.period_range[-1]` (last period of the full dataset) where `Evaluation.create` correctly uses `train_set.period_range[-1]`. `train_set` is now in scope right above, so this is a one-word fix.

Also note `test_backtest_uses_n_test_sets_for_retraining` replaces the old stride test, so stride propagation across the three new call sites is no longer covered.

PR #568, opened 2026-09-08
feat: pluggable future-weather providers with look-ahead flag

knutdrand:
Implements CLIM-1042. Supersedes #566, which contained only a design document; that document is not part of this branch, so #566 can be closed.

## What changes

Future-weather providers become plugins in the same sense as metrics, backtest plots and threshold strategies: a class with an `id`, `name` and `description`, registered via `@weather_provider` and resolved by id from a registry.

Providers declare `leaks_future_data`. The call site passes the forecast window's observations **only** to providers that declare it:

```python
provider_cls = resolve_weather_provider(provider_id)
provider_cls().get_future_weather(
    historical_data,
    period_range,
    future_data=future_data if provider_cls.leaks_future_data else None,
    params=params,
)
```

A provider that does not declare the flag is structurally unable to look ahead, rather than trusted not to. "Use the real weather" is now the `observed` provider instead of the *absence* of a provider, which is what made the leak reachable by omission.

Forecasting providers only see seasonal numeric covariates. `get_future_weather` strips the target, string columns (org unit parents, codes) and non-seasonal numerics such as `population` before calling the provider, and reattaches them carried forward from their last observation. Regressing population on month-of-year would otherwise hand the model a climatological population instead of the current one. Providers stay pure forecasters and none has to re-implement that rule.

## Providers

| id | leaks | notes |
|---|---|---|
| `climatology` | no | per-location regression on month/week-of-year, the previous `QuickForecastFetcher` behaviour. **Default.** |
| `damped_persistence` | no | climatology plus the current anomaly, damped by `phi**h`; `phi` defaults to the lag-1 autocorrelation of the anomaly series |
| `mstl_arima` | no | MSTL decomposition + AICc-selected ARIMA; needs more than 2 seasonal cycles |
| `observed` | **yes** | perfect foresight, diagnostic runs only; cannot be used to predict ahead |

`future_weather_provider` is added to `BacktestParams` and `PredictionParams`, validated against the registry and threaded through `chap eval`, the REST API and the worker functions, so evaluation and prediction use the same source. `PredictionParams` additionally rejects providers that declare `leaks_future_data`, so a request for `observed` fails with 422 at every prediction entry point. A prediction setup inherits the provider its backtest was evaluated with. `GET /v1/analytics/weather-providers` lists the providers with their look-ahead flag.

The backtest row is the single source of truth for the provider. `Evaluation` reads it from the row, the worker persists it before the backtest is constructed, and NetCDF export and database persistence cannot disagree.

## Breaking change

**The default is `climatology` everywhere.** `chap eval` previously received the observed weather, so re-running an existing evaluation will produce different numbers. Pass `--backtest-params.future-weather-provider observed` to restore the old behaviour. The REST paths already used climatology and are unaffected.

Routing `chap eval` through climatology also means every CLI evaluation now fits `MonthlyClimatePredictor`. Its NaN assertion has been replaced by fitting on the observed rows of each covariate, so a gap in rainfall no longer aborts the run; a covariate that is entirely missing raises a ValueError naming the column. Data at resolutions other than month or week raises a ValueError pointing at the `observed` provider instead of a bare assertion.

## `.nc` backward compatibility

The provider is written into the evaluation `.nc` file. Files without the attribute are read back as `observed` (`LEGACY_WEATHER_PROVIDER_ID`): that is what produced them, and without the hedge they would silently inherit today's default and misreport how they were run.

## How the providers actually perform

Rolling-origin backtest, 3 periods ahead, MAE vs `climatology` over four real datasets (Vietnam/Laos admin1 monthly, Malawi district monthly, Laos weekly):

| | rainfall | mean_temperature |
|---|---|---|
| `damped_persistence` | +0.0%, -0.1%, +2.9%, -0.4% | +4.7%, +6.2%, +5.7%, -2.4% |
| `mstl_arima` | -1.1%, -4.8%, -18.7%, -21.1% | +14.1%, +2.6%, +1.9%, +15.3% |

`damped_persistence` never loses more than 2.4% and costs less than climatology to fit; when anomalies carry no signal its damping factor goes to ~0 and it degrades into climatology. `mstl_arima` is a strong choice for temperature-driven models, especially weekly, where climatology's 52 week-of-year dummies overfit, but is the wrong model for rainfall, which is intermittent and right-skewed, and it is 100-300x slower.

## New dependency

`statsmodels`, for `mstl_arima`. Pure Python over numpy/scipy; avoids the numba toolchain that `statsforecast` pulls in.

## Not included

`imported` (user-supplied forecast datasets) and a real seasonal-forecast product both need a decision on how the forecast data is referenced. Nothing yet asserts that a prediction's provider matches the backtest's beyond the setup-run path, which inherits it.

## Testing

`make lint` clean (ruff, mypy, pyright). Full suite: 1437 passed, 122 skipped.

### DIFF
--- a/chap_core/assessment/evaluation.py
+++ b/chap_core/assessment/evaluation.py
@@ -19,6 +19,11 @@ import pandera.pandas as pa
 import xarray as xr
 from packaging.version import Version
 
+from chap_core.assessment.weather_providers import (
+    DEFAULT_WEATHER_PROVIDER_ID,
+    LEGACY_WEATHER_PROVIDER_ID,
+)
+
 if TYPE_CHECKING:
     from chap_core.api_types import BacktestParams
 
@@ -96,6 +101,7 @@ def _flat_data_to_xarray(flat_data: "FlatEvaluationData", model_metadata: dict) 
         "split_periods": json.dumps(model_metadata.get("split_periods", [])),
         "org_units": json.dumps(model_metadata.get("org_units", [])),
         "historical_context_periods": model_metadata.get("historical_context_periods", 0),
+        "future_weather_provider": model_metadata.get("future_weather_provider", DEFAULT_WEATHER_PROVIDER_ID),
         "chap_version": CHAP_VERSION,
     }
 
@@ -243,6 +249,7 @@ class EvaluationBase(ABC):
         info: "BacktestCreate",
         historical_observations: list[Observation] | None = None,
         historical_context_periods: int = 0,
+        future_weather_provider: str = DEFAULT_WEATHER_PROVIDER_ID,
     ) -> "EvaluationBase": ...
 
 
@@ -259,6 +266,7 @@ class Evaluation(EvaluationBase):
         backtest: "Backtest",
         historical_observations: list[Observation] | None = None,
         historical_context_periods: int = 0,
+        future_weather_provider: str = DEFAULT_WEATHER_PROVIDER_ID,
     ):
         """
         Initialize Evaluation with a Backtest object.
@@ -268,11 +276,19 @@ class Evaluation(EvaluationBase):
             historical_observations: Optional list of Observation objects for historical
                 context (periods before split points, for plotting)
             historical_context_periods: Number of periods of historical context stored
+            future_weather_provider: Id of the provider that supplied the climate
+                covariates for each forecast window
         """
         self._backtest = backtest
         self._historical_observations = historical_observations or []
         self._historical_context_periods = historical_context_periods
+        self._future_weather_provider = future_weather_provider
         self._flat_data_cache: FlatEvaluationData | None = None
+
+    @property
+    def future_weather_provider(self) -> str:
+        """Id of the future-weather provider these results were produced with."""
+        return self._future_weather_provider
 
     @classmethod
     def from_backtest(cls, backtest: "Backtest") -> "Evaluation":
@@ -296,6 +312,7 @@ class Evaluation(EvaluationBase):
         info: BacktestCreate,
         historical_observations: list[Observation] | None = None,
         historical_context_periods: int = 0,
+        future_weather_provider: str = DEFAULT_WEATHER_PROVIDER_ID,
     ) -> "Evaluation":
         info.created = datetime.datetime.now()
         backtest = Backtest(
@@ -370,6 +387,7 @@ class Evaluation(EvaluationBase):
             backtest,
             historical_observations=historical_observations,
             historical_context_periods=historical_context_periods,
+            future_weather_provider=future_weather_provider,
         )
 
     @classmethod
@@ -411,6 +429,7 @@ class Evaluation(EvaluationBase):
             prediction_length=backtest_params.n_periods,
             n_test_sets=backtest_params.n_splits,
             stride=backtest_params.stride,
+            future_weather_provider=backtest_params.future_weather_provider,
         )
 
         # Run backtest
@@ -433,6 +452,7 @@ class Evaluation(EvaluationBase):
             n_splits=backtest_params.n_splits,
             stride=backtest_params.stride,
             n_retrain=backtest_params.n_retrain,
+            future_weather_provider=backtest_params.future_weather_provider,
         )
 
         # Calculate number of periods based on dataset period type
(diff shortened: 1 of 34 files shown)


mortenoh:
Found three issues:

1. **[P1] REST backtests save the wrong provider.** The worker uses `future_weather_provider` for forecasting but never assigns it to `info`, whose default remains `climatology`. Both REST entry points therefore save non-default runs—including `observed`—as climatology, hiding look-ahead usage. Persist the resolved provider before constructing the backtest. [Code](https://github.com/dhis2-chap/chap-core/blob/ad7aad50aa1d248c52c17d757329b1be37debc73/chap_core/rest_api/db_worker_functions.py#L125-L145)

2. **[P2] Evaluation conversions lose provider metadata.** `Evaluation.from_backtest()` ignores the backtest’s provider and takes the constructor default. Reproduced: an `observed` backtest becomes a `climatology` evaluation, which then exports incorrect `.nc` metadata. Conversely, `from_file()` reads the provider without setting it on the underlying backtest. Keep these representations consistent. [Code](https://github.com/dhis2-chap/chap-core/blob/ad7aad50aa1d248c52c17d757329b1be37debc73/chap_core/assessment/evaluation.py#L282-L304)

3. **[P2] MSTL crashes at the advertised minimum history.** Exactly 24 monthly observations pass the length check, but MSTL drops seasonal periods equal to half the series length and raises `UnboundLocalError`. Reproduced with statsmodels 0.15.0; 25 observations succeed. Handle this boundary or require more than two cycles. [Code](https://github.com/dhis2-chap/chap-core/blob/ad7aad50aa1d248c52c17d757329b1be37debc73/chap_core/assessment/weather_providers/mstl_arima.py#L122-L134)

Validation: 33 focused tests passed, plus separate reproductions above. Full suite not run.

chap_core/assessment/evaluation.py
```diff
@@ -268,12 +276,20 @@ def __init__(
             historical_observations: Optional list of Observation objects for historical
                 context (periods before split points, for plotting)
             historical_context_periods: Number of periods of historical context stored
+            future_weather_provider: Id of the provider that supplied the climate
+                covariates for each forecast window
         """
         self._backtest = backtest
         self._historical_observations = historical_observations or []
         self._historical_context_periods = historical_context_periods
+        self._future_weather_provider = future_weather_provider
```
mortenoh: **P2 — Keep the evaluation provider and backtest row consistent.** The provider is stored separately from the underlying row, so `Evaluation(backtest, future_weather_provider="observed")` leaves `backtest.future_weather_provider` unchanged. I reproduced an observed evaluation reverting to climatology after `to_backtest()` → `from_backtest()`. Conversely, `Evaluation(backtest)` labels an observed row as climatology. The explicit override in `from_samples_with_truth()` also never updates the constructed row. Please use the row as the source of truth and apply explicit overrides there, so database persistence and NetCDF exports agree.

chap_core/rest_api/data_models.py
```diff
@@ -66,6 +67,19 @@ class PredictionParams(DBModel):
 
     model_id: str = Field(description="Canonical name of the configured model to run.")
     n_periods: int = Field(default=3, gt=0, description="Number of future periods to forecast.")
+    future_weather_provider: str = Field(
```
mortenoh, line 70: **P2 — Thread provider selection through saved prediction setups.** This adds selection for the direct prediction endpoint, but `run_prediction_setup()` in `crud.py:1071` still constructs `PredictionParams(model_id=model_id, n_periods=request.n_periods)`, which always selects climatology. `RunPredictionSetupRequest` has no provider field and forbids extra fields, so callers cannot override it. Promoting a damped_persistence or mstl_arima backtest into a setup therefore runs predictions with a different weather source, with no way to select the evaluated provider. Please expose and forward the selection on the setup run path, or inherit it from the setup's backtest.

mortenoh (commented):
Found two issues with provider persistence and prediction-setup routing. Validation: 36 focused tests passed; MSTL/ARIMA tests were not run because the local environment lacks statsmodels.

mortenoh:
Left as is in the review fixes: the `params` argument on `get_future_weather` stays. It is only reachable from tests, since neither `BacktestParams` nor `PredictionParams` exposes it, so `damping` and `order` cannot be set from the CLI or API. Removing it, or threading it through the params models, is a design call for the author.

mortenoh (approved):

PR #569, opened 2026-09-08
feat: persist backtest parameters and consolidate the backtest endpoints

ivargr:
## Summary

Backtest rows now store the four evaluation parameters (`n_periods`, `n_splits`, `stride`, `n_retrain`) as the values that actually ran, exposed on `BacktestRead`. `n_retrain` was silently dropped on both REST paths and always ran as 1; it is now threaded through `run_backtest`. `POST /v1/crud/backtests` is removed in favour of `POST /v1/analytics/create-backtest`, which now also accepts the configured model's integer id, so there is one endpoint and one set of defaults for backtesting a stored dataset.

Jira: CLIM-1052 (under CLIM-980, benchmarking).

## Changes

- `BacktestBase` inherits `BacktestParams`, so `Backtest` gets four NOT NULL columns and `BacktestRead` / `BacktestCreate` carry the same fields. `BacktestParams` stays the single definition and source of defaults.
- `run_backtest` accepts `n_retrain`, passes it to the evaluator, drops its competing `n_splits=10` default, and writes the resolved values onto the row after the `ewars_plus` and derive-from-dataset overrides. `run_backtest_from_dataset` forwards `n_retrain`; its weekly `stride=4` override already runs before the call, so the stored stride is the resolved one.
- CLI evaluations (`Evaluation.create`) and the naive simulator record their parameters as well.
- Alembic revision `d0e1f2a3b4c5`: adds the columns and reconstructs legacy rows from their forecasts (`n_splits` from distinct split points, `n_periods` from periods per split, `stride` from the smallest gap between splits, `n_retrain=1` since the REST path never forwarded it). Handles the generic startup migration having already created the columns with `0`. Rows without forecasts get the `BacktestParams` defaults with a warning.
- `MakeBacktestRequest.model_id` is `int | str`; `crud/backtests` removed; docs and the integration flow test updated.

## Notes for review

- Checked `dhis2-chap/chap-frontend` at `44aa78b` (2026-09-08): `POST /v1/crud/backtests` appears only as the generated client method `createBacktestV1CrudBacktestsPost` with no callers; the modeling app uses `create-backtest-with-data/` only, and the e2e fixtures only GET `/v1/crud/backtests/{id}/info`, which stays. Regenerating the client will drop the dead method. Job-type strings are unchanged.
- API and worker are separate images. `run_backtest` gains a keyword argument and `BacktestCreate` gains fields in the pickled task payload, so workers should roll before API pods.
- `create-backtest` still does not apply the weekly `stride=4` bump that `create-backtest-with-data` does; pre-existing and left as is.

## Test plan

- New: endpoint forwards all four params and accepts an integer `modelId`; `run_backtest` persists resolved values (`n_periods=None` resolves to 3 on monthly data, integer model id resolves to name); `n_retrain=2` retrains twice; `nRetrain=0` is rejected; migration derives stride across month and 53-week ISO year boundaries; docker-backed migration test backfills a legacy backtest to `(3, 2, 2, 1)` and one without forecasts to the defaults.
- `make lint` passes. `make test` passes except `tests/test_celery.py` and `tests/integration/rest_api/test_jobs_routes.py`, which fail at collection on this machine's redis server (pre-existing, identical on master).

### DIFF
--- a/alembic/versions/d0e1f2a3b4c5_add_backtest_params.py
+++ b/alembic/versions/d0e1f2a3b4c5_add_backtest_params.py
@@ -1,0 +1,119 @@
+"""add_backtest_params
+
+Store the evaluation parameters a backtest ran with (n_periods, n_splits, stride,
+n_retrain) on the backtest row. Legacy rows are reconstructed from their forecasts:
+one split per distinct last_seen_period, n_periods from the periods forecast per
+split, stride from the smallest gap between splits. n_retrain is 1 for every legacy
+row because the REST path never forwarded it.
+
+Revision ID: d0e1f2a3b4c5
+Revises: c9d0e1f2a3b4
+Create Date: 2026-09-08
+
+"""
+
+import logging
+from collections import defaultdict
+from collections.abc import Sequence
+from datetime import datetime
+
+import sqlalchemy as sa
+
+from alembic import op
+
+revision: str = "d0e1f2a3b4c5"
+down_revision: str | Sequence[str] | None = "c9d0e1f2a3b4"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+logger = logging.getLogger("alembic.runtime.migration")
+
+PARAM_COLUMNS = ("n_periods", "n_splits", "stride", "n_retrain")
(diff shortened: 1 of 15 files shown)


knutdrand (approved):
very good

PR #572, opened 2026-09-09
refactor: extract backtest evaluation details into a BacktestSpecification

ivargr:
Implements CLIM-1070, under the benchmarking epic CLIM-980.

## Why

The parameters that decide whether two backtests are comparable — `n_periods`, `n_splits`, `stride`, `n_retrain`, the dataset, and the resulting org unit set — lived as columns on each backtest row. They were owned by a single backtest and could not be referred to by anything else.

Benchmarking needs the opposite. Comparing models means running them against the same specification, and adding a model to an existing comparison later has to be a new backtest pointing at a specification that already exists, not a re-run of everything. While the details are duplicated per row, "these results are comparable" is something every caller has to verify by comparing columns.

## What changes

`BacktestSpecification` holds the dataset, the evaluation parameters and the resolved org unit set. Rows are deduplicated and immutable; `Backtest` gains a not-null `specification_id` and loses its own copies of the four parameter columns.

The uniqueness key is **derived from `BacktestParams` rather than listed**:

```python
__table_args__ = (
    UniqueConstraint("dataset_id", *BacktestParams.model_fields, name="uq_backtestspecification_params"),
)
```

A parameter added to `BacktestParams` joins the key automatically instead of silently letting two different setups share a row. A test pins that invariant. Widening the constraint in the database still needs a migration.

`org_units` is stored but is **not** part of the key, because it is derived from it. `validate_and_filter_dataset_for_evaluation` drops locations whose target is empty across the training window, and the length of that window is `n_periods + (n_splits - 1) * stride`. The same dataset under different parameters yields different org units, so they belong to the specification rather than to the dataset.

Resolve-or-create lives in `SessionWrapper.get_or_create_backtest_specification`. Concurrent backtest jobs can race, so a unique violation re-reads the winner's row rather than failing the backtest.

Not in scope: a benchmark table. A benchmark is a specification with more than one backtest under it, which is a query rather than a table.

## REST contract

`BacktestRead` is unchanged. It keeps the four parameters as flat fields, now read through the relationship, and its generated schema is byte-identical — a test pins the property set, and the read paths eager-load the specification so listings do not fall into an N+1.

One endpoint does change. `GET /v1/crud/backtests/{id}/full` declares `response_model=Backtest`, the table class itself, so its component gains `specificationId` and the four parameters become `readOnly` and `required` instead of optional-with-a-default. Nothing is removed or renamed, and for a response-only schema this is a tightening, but it is a diff rather than literally no change.

## Migration

`f2a3b4c5d6e1` groups existing backtests by `(dataset_id, parameters)` into one specification each, points the rows at them, then drops the columns. It handles the startup path where `create_all` has already made the table and the generic metadata migration has already added `specification_id` filled with `0`.

Two things about the backfilled data cannot be repaired, and both are recorded in the migration docstring so migrated rows are not mistaken for properly resolved ones:

- **Parameters on pre-existing rows were themselves reconstructed** from forecasts by #569. A split whose predictor returned nothing left no forecasts and is invisible, so such a backtest was reconstructed with a lower `n_splits` and a wider `stride`. Two backtests that really ran the same setup can carry different parameters and will land on different specifications. Legacy grouping is fragmented, and fragmented specifically for the models that misbehaved.
- **Copied `org_units` are output-derived**, accumulated from what each model produced rather than from the filtered input, so a model that dropped locations under-reports. The union across the backtests sharing a specification is the best available estimate: an org unit any of them forecast must have been in the input. Specifications created from now on resolve `org_units` correctly.

`downgrade()` copies the parameters back and is lossy: a specification no backtest points at is dropped with the table.

## Tests

`tests/test_alembic_migrations.py` seeds a third legacy backtest whose forecasts match `legacy_backtest`, so it derives identical parameters and must land on the same specification. It then asserts the three backtests resolve to exactly two specifications, that the parameters read through the link are what the old assertions expected, that `org_units` is the union across the contributing rows, that the four columns are gone and that `specification_id` is not null.

New behaviour tests cover identical parameters sharing one specification, differing parameters and differing datasets producing different ones, `org_units` matching the post-filter locations including a case where the filter drops one, and the read contract.

## The future-weather provider

#568 merged first, and this branch is rebased on top of it. `future_weather_provider` moves off `backtest` onto the specification with the rest of the parameters: two backtests given different climate covariates for their forecast windows are not comparable, so keeping the provider on the row would have left the one parameter that matters most for comparability outside the thing that certifies it.

The model side needed no edit — the key is derived from `BacktestParams.model_fields`, so the field joined it on its own, and a test asserts a `climatology` and an `observed` run land on different specifications. `Backtest.future_weather_provider` became a fifth read-through property, so `Evaluation.future_weather_provider` and the prediction-setup inheritance in `crud.py` keep working unchanged.

The migration takes `down_revision = "e1f2a3b4c5d6"` and carries a type per parameter column rather than assuming integers, since the provider is a VARCHAR. It moves all five columns, guarding each drop so a database missing some of them still ends up with none on `backtest`.

Tests that set the provider by assigning to the backtest now go through the specification. Two of them did so on a *persisted* backtest, where mutating a shared immutable specification would have changed the provider for every backtest pointing at it; those point the backtest at a new specification instead.

## Out of scope

The benchmark batch endpoint, covariate mapping, optional covariates, HPO run persistence and marketplace lookup stay in their own CLIM-980 issues. The `ewars_plus` `n_periods=3` override and the weekly `stride=4` bump are tracked separately, but note the override mutates a key field, so it has to be applied before the specification is resolved — as it is here.

### DIFF
--- a/alembic/versions/f2a3b4c5d6e1_extract_backtest_specification.py
+++ b/alembic/versions/f2a3b4c5d6e1_extract_backtest_specification.py
@@ -1,0 +1,192 @@
+"""extract_backtest_specification
+
+Move the evaluation parameters (n_periods, n_splits, stride, n_retrain,
+future_weather_provider) off the backtest row and onto a deduplicated
+``backtestspecification`` row that backtests point at, so that two backtests sharing a
+specification are comparable by construction.
+
+Existing rows are grouped by (dataset_id, and every one of those parameters) into one
+specification each, and the specification's org_units is the union of the org_units of
+the backtests that map to it. The provider is part of the key because two backtests
+given different climate covariates for their forecast windows are not comparable, even
+when every other parameter matches.
+
+Two things about the migrated data are worth knowing, because neither can be
+recovered here:
+
+- The parameters of rows predating d0e1f2a3b4c5 were reconstructed from their
+  forecasts rather than recorded. A split whose predictor returned nothing left no
+  forecasts and is therefore invisible, so such a backtest was reconstructed with a
+  lower n_splits and a wider stride. Two backtests that genuinely ran the same setup
+  can carry different parameters and will land on different specifications. Legacy
+  grouping is fragmented, and fragmented specifically for the models that misbehaved.
+- backtest.org_units is accumulated from what the model produced, not from the
+  filtered input, so for a model that dropped org units the copied set under-reports.
+  The union across the backtests sharing a specification is the best estimate
+  available: an org unit any of them forecast must have been in the input.
+  Specifications created from now on resolve org_units from the filtered dataset.
+
+Revision ID: f2a3b4c5d6e1
+Revises: e1f2a3b4c5d6
+Create Date: 2026-09-09
+
+"""
+
+import json
+import logging
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+
+from alembic import op
+
+revision: str = "f2a3b4c5d6e1"
+down_revision: str | Sequence[str] | None = "e1f2a3b4c5d6"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+logger = logging.getLogger("alembic.runtime.migration")
+
+# The parameters as of this revision, with the SQL type each column carries. A parameter
+# added to BacktestParams later joins the specification's uniqueness key automatically,
+# but still needs its own migration to move the column and rebuild the constraint; this
+# list stays frozen at what was here.
+PARAM_TYPES = {
+    "n_periods": sa.Integer(),
+    "n_splits": sa.Integer(),
+    "stride": sa.Integer(),
+    "n_retrain": sa.Integer(),
+    "future_weather_provider": sa.String(),
+}
+PARAM_COLUMNS = tuple(PARAM_TYPES)
+# BacktestParams defaults when this migration was written. Only reached on downgrade,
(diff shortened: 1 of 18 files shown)


mortenoh:
Found one regression:

**[P2] Unused specifications prevent dataset deletion** — [tables.py, lines 29–32](https://github.com/dhis2-chap/chap-core/blob/4dfc7a6ce05236b129287aa96210e69f45c1b4d1/chap_core/database/tables.py#L29-L32).

Deleting a backtest leaves its specification behind, but the new dataset foreign key has no cleanup or cascade handling. Consequently, deleting the dataset fails even after all its backtests are removed. Failed runs also leave specifications because creation commits before evaluation starts.

I reproduced the foreign-key failure with SQLite enforcement enabled. Please remove unused specifications during dataset deletion, or add appropriate cascade handling in both the ORM and migration, with a regression test covering deletion after the last backtest is removed.

Validation: 45 focused tests passed. PostgreSQL migration tests were not run.

mortenoh:
The regression test builds its DataSet inline instead of using a fixture, which rule 12 discourages. It also tests the orphan-specification case directly rather than the literal "delete last backtest, then dataset" flow. Both are minor, and the underlying cause is covered.

mortenoh (approved):

PR #583, opened 2026-09-10
fix(validate): reject non-numeric required covariates and non-finite values

mortenoh:
## Problem

`chap validate` reports success, or only a warning, for datasets that then fail at run time.

- A required covariate that holds strings produces the warning `Column 'x' is non-numeric and will be ignored by models` and no error. The column is not ignored: `DataSet.to_csv` keeps it and the external model runner writes it to `training_data.csv` unchanged, so the model receives text where it expects numbers.
- A covariate containing `inf` or `-inf` produces no issue at all, because the missing-value check only looks for NaN. The climatology weather provider then fails inside sklearn with `Input y contains infinity or a value too large for dtype('float64')`.

Validation is only run by the `chap validate` command, so the point of the command is to catch exactly these cases before a backtest is started.

## Overview

Changes are limited to `chap_core/services/dataset_validation.py`:

- The missing-value check now uses `np.isfinite`, so `inf` and `-inf` are reported as errors alongside NaN, with the same per-location and per-period detail. Message: `Missing or non-finite values in '<field>'`.
- The required-covariate check now also verifies the dtype of each required covariate that is present. A non-numeric one is an error: `Required covariate '<field>' is non-numeric`.
- The non-numeric warning for columns the model does not require now says what actually happens: `Column '<field>' is non-numeric and is passed to the model as-is`. It stays a warning, since a stray text column that the model never reads is not fatal.

## Examples

`example_data/laos_subset.csv` with an added string column `region_class`, validated against a model that requires it:

Before:

```
Warnings (1):
  - Column 'region_class' is non-numeric and will be ignored by models
```

After:

```
Warnings (1):
  - Column 'region_class' is non-numeric and is passed to the model as-is

Errors (1):
  - Required covariate 'region_class' is non-numeric
```

`example_data/laos_subset.csv` with one `rainfall` value set to `inf`:

Before:

```
Validation passed: no issues found.
```

After:

```
Errors (1):
  - Missing or non-finite values in 'rainfall'
    Location: Bokeo
    Time periods: 2010-01
```

## Verification

- New tests in `tests/test_validate_cli.py`: `test_validate_errors_on_non_numeric_required_covariate` and `test_validate_errors_on_infinite_covariate`, both built on the existing `LAOS_SUBSET` fixture pattern.
- The existing `test_validate_warns_on_non_numeric_field` still passes: a string column the model does not require remains a warning.
- `make lint` and `make test` pass (1482 passed, 122 skipped).

### DIFF
--- a/chap_core/services/dataset_validation.py
+++ b/chap_core/services/dataset_validation.py
@@ -80,21 +80,21 @@ def _check_nan_covariates(dataset: DataSet) -> list[ValidationIssue]:
                     issues.append(
                         ValidationIssue(
                             level="warning",
-                            message=f"Column '{field_name}' is non-numeric and will be ignored by models",
+                            message=f"Column '{field_name}' is non-numeric and is passed to the model as-is",
                         )
                     )
                 continue
             if not np.issubdtype(values.dtype, np.floating):
                 continue
-            isnan = np.isnan(values)
-            if np.any(isnan):
-                nan_periods = [data.time_period[i].to_string() for i in np.flatnonzero(isnan)]
+            invalid = ~np.isfinite(values)
+            if np.any(invalid):
+                invalid_periods = [data.time_period[i].to_string() for i in np.flatnonzero(invalid)]
                 issues.append(
                     ValidationIssue(
                         level="error",
-                        message=f"Missing values in '{field_name}'",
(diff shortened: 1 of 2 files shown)


knutdrand (approved):
Very good

PR #584, opened 2026-09-10
fix: remove demonstration metrics from evaluation results (CLIM-1086)

edvinstava:
Removes the `example_metric` and `sample_count` metrics. Both were development examples, but were registered in the metric registry, so every evaluation computed, stored and returned them next to real scores such as CRPS and MAE.

- delete `example_metric.py` and `test_metrics.py` and their registry entries/exports
- drop the sample count line from the generated model card
- drop the stale sample count row from the export-metrics docs


knutdrand (approved):

mortenoh:
The removal itself is clean: no stale references to the deleted metrics remain in code, docs, configs, or tests, and `tests/evaluation` passes on the PR head. Three things worth fixing before merge.

**1. The docs table this PR edits is still wrong.** `docs/chap-cli/evaluation-workflow.md` drops the `test_sample_count` row but leaves four other ids that no metric produces:

| In docs table | Actual metric id |
|---|---|
| `rmse_aggregate` | `rmse` |
| `mae_aggregate` | `mae` |
| `ratio_within_10th_90th` | `coverage_10_90` |
| `ratio_within_25th_75th` | `coverage_25_75` |

The "Selecting Specific Metrics" snippet a few lines below already uses `rmse` and `mae`, so the page contradicts itself, and passing a table id to `chap export-metrics` raises `Invalid metric IDs`. Since the point of the PR is making the docs match the registry, the table and the CSV example above it should use the real ids. Ideally also list the other registered ids (`crps_norm`, `crps_log1p`, `mape`, `ratio_above_truth`, `winkler_score_*`, `sensitivity`, `specificity`, `outbreak_accuracy`, `peak_value_diff`, `peak_period_lag`).

**2. The rewritten tests lost their only value assertion.** In `tests/evaluation/test_metrics.py`, the old `metrics["sample_count"] == 24.0` pinned the row count flowing through the backtest -> flat-data pipeline. The replacement `"crps" in metrics` is true even when the value is `None` or `NaN`. The added `assert not {"sample_count", "example_metric"} & set(...)` checks can only fail if someone re-adds those exact modules, so they are dead weight that keeps the deleted names alive in grep results. Suggest asserting `metrics["crps"]` is a finite float and dropping the negative checks (or comparing against an explicit expected-id set).

**3. Two redundant registration paths in `chap_core/assessment/metrics/__init__.py`.** `_discover_metrics()` imports a hand-maintained module list, but the explicit class-import block right below it already imports every module and triggers the `@metric()` decorators. The lists are already out of sync: `peak_diff` is registered only via the import block. The comment "Import metric classes for backwards compatibility in exports" is misleading since that block is what actually registers metrics. This predates the PR, but since the PR edits both lists it is a good moment to delete `_discover_metrics()` and fix the comment.

Minor, non-blocking:

- Backtests persisted before this change still carry a `sample_count` key in `Backtest.aggregate_metrics`. A client that iterates those keys and requests `/metric-plots/.../sample_count` now gets a 400. Probably fine for a demo metric, but worth a note in the description.
- `ALL_METRIC_FACTORIES` in the test file is a fourth hand-copied list of metric classes. Deriving it from `available_metrics` would make future removals a one-file change.

PR #588, opened 2026-09-10
feat(cli): install, update and uninstall marketplace models

edvinstava:
Adds `chap install`, `chap update` and `chap uninstall` for chapkit models from the [model marketplace](https://github.com/dhis2-chap/model-marketplace).

## Decisions

- **Only verified stable versions install.** The `channels.stable` pin must have `status: verified`; `latest`, unverified versions and templates are rejected. A `sha-` tag must be a prefix of the pinned commit.
- **Unreviewed code needs `--accept-risk` every time.** Custom images (`--image`) and models from a non-default registry (`CHAP_MARKETPLACE_URL`) require it on both install and update, with a warning about running their code, sharing data and using forecasts.
- **Models live in a separate Compose overlay.** Deployments get `compose.marketplace.yml` next to the base file; `--local` uses `~/.chap/compose.models.yml` under its own project name and prints a `127.0.0.1` URL for `chap eval`. The overlay is only replaced after the pull and start succeed; a failed update restores the previous image.
- **Updates keep operator settings and data.** The previous service definition and data volume are reused. `DATABASE_URL` is pinned to the mounted volume on install and update so models whose `WORKDIR` is not `/app` start on the read-only filesystem. `--platform` is kept for later updates, and a failed pull suggests `--platform linux/amd64`.
- **Uninstall keeps data unless asked.** The container is removed and the service dropped from the overlay; the data volume stays unless `--delete-data` is passed. The overlay is deleted when the last model goes.

[CLIM-1085](https://dhis2.atlassian.net/browse/CLIM-1085)

[CLIM-1085]: https://dhis2.atlassian.net/browse/CLIM-1085?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/chap_core/cli.py
+++ b/chap_core/cli.py
@@ -19,6 +19,7 @@ from chap_core.cli_endpoints import (
     explain,
     forecast,
     generate_modelcard,
+    marketplace,
     model,
     preference_learn,
     report,
@@ -42,6 +43,7 @@ evaluate.register_commands(app)
 explain.register_commands(app)
 forecast.register_commands(app)
 model.register_commands(app)
+marketplace.register_commands(app)
 preference_learn.register_commands(app)
 report.register_commands(app)
 utils.register_commands(app)
--- a/chap_core/cli_endpoints/marketplace.py
+++ b/chap_core/cli_endpoints/marketplace.py
@@ -1,0 +1,193 @@
+"""Install and update chapkit model services in a CHAP Compose deployment."""
+
+import logging
+import re
+import subprocess
+import tempfile
+from pathlib import Path
+from typing import Annotated
+
+from cyclopts import Parameter
+
+logger = logging.getLogger(__name__)
+
+ModelArg = Annotated[str, Parameter(help="Marketplace model ID, or a name for a custom chapkit model.")]
+ComposeArg = Annotated[
+    tuple[Path, ...],
+    Parameter(help="Base Compose files in deployment order. Defaults to compose.yml."),
+]
+ImageArg = Annotated[str | None, Parameter(help="Custom chapkit container image (requires --accept-risk).")]
+RiskArg = Annotated[
+    bool,
+    Parameter(help="Accept responsibility for running unreviewed custom model code and using its forecasts."),
+]
+LocalArg = Annotated[
+    bool, Parameter(help="Run a standalone local service for CLI evaluations (no CHAP server needed).")
+]
+PlatformArg = Annotated[str | None, Parameter(help="Container platform, e.g. linux/amd64 for R-INLA on Apple Silicon.")]
+
+
+def install(
+    model: ModelArg,
+    *,
+    compose_file: ComposeArg = (Path("compose.yml"),),
+    image: ImageArg = None,
+    accept_risk: RiskArg = False,
+    local: LocalArg = False,
+    platform: PlatformArg = None,
+) -> None:
+    """Install a marketplace model's verified stable version into CHAP.
+
+    Run from your CHAP Docker Compose deployment directory. Custom chapkit images
+    can be installed with --image IMAGE --accept-risk. Docker Compose is required.
+    """
+    _deploy(model, compose_file, image, accept_risk, local, platform, updating=False)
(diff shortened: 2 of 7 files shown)


mortenoh:
### Should fix before merge

1. **Data volume at `/app/data` is probably not writable** (`chap_core/cli_endpoints/marketplace.py:15-19`, `:222`).
   chapkit 2.x templates use `WORKDIR /work` and only create and chown `/work/data` for the `chapkit` user
   (uid 1000). If `/app/data` does not exist in the image, Docker creates the named volume owned by root,
   and SQLite cannot create `chapkit.db`, so `up --wait` fails and the install rolls back.
   Please check this against a real chapkit-py image. A longer-term fix would be an absolute default data
   path in chapkit, or a data-path field in the marketplace schema.
2. **Rollback does not restore the old image when the image reference is unchanged** (`marketplace.py:262`).
   `pull` has already moved the local tag, so `up` with the old overlay recreates the same broken image.
   This is the normal case for `chap update <custom> --accept-risk`. One fix is to record the old image ID
   (`docker image inspect`) before pulling and restore from that ID.
3. **The stable-pin check does not reject mutable tags** (`chap_core/services/model_marketplace.py:76`).
   For a non-`sha-` tag, `required_prefix` is the commit itself, so the commit check always passes. Only the
   exact string `latest` is rejected, so tags like `main`, `edge` or `v1` count as verified pins, and so does
   a one-character prefix like `sha-5`. Suggestion: require `sha-` tags with a minimum prefix length (e.g. 7),
   or pin by digest.
4. **The source registry is not saved with the model** (`marketplace.py:187`).
   `update` resolves the model against the `CHAP_MARKETPLACE_URL` in the current shell. A model installed from an
   org registry can be silently replaced by a default-marketplace model with the same ID, and the risk gate
   is skipped. Suggestion: store the registry in the service (e.g. `x-chap-registry`) and use it on update.
5. **Explicit `-f` flags stop Compose from loading `compose.override.yml`** (`marketplace.py:148`, plus the
   docs' suggested command). Deployments that use an override (the repo ships
   `compose.override.yml.example`) can end up with orphaned services or a different project name or
   network. With `--remove-orphans`, those orphaned services are removed.

### Robustness

6. The overlay is published from a `NamedTemporaryFile`, so `compose.marketplace.yml` always ends up with mode 0600 (`:157`).
   Other operators and service accounts can no longer read it.
7. Only `CalledProcessError` triggers rollback (`:259`). Ctrl-C during the 120 s `--wait` can leave the container on the new image while the overlay still names the old one.
8. Uninstalling the last model deletes the overlay (`:112`), but the CLI told operators to always pass `-f compose.marketplace.yml`, so those commands then fail. Suggestion: write `services: {}` instead.
9. `dict(service.get("environment") or {})` breaks on the list form of Compose `environment` (`:231`).
10. `version` from the registry and `--platform` are written to the overlay unescaped (`:234`). A `$` in either breaks every later Compose command.
11. `--delete-data` runs `docker volume rm` after the overlay has been pruned (`:116`). If that fails, the command cannot be retried ("not installed").
12. A `volumes:` key or service entry that is null in a hand-edited overlay causes a raw TypeError or AttributeError (`:139`).
13. The pull-failure hint suggests `--platform linux/amd64` even when the saved service already has that platform (`:249`).
14. `compose config` and `compose port` use `capture_output=True` and `check=True` (`:100`, `:271`), so Docker's stderr is lost from the error.

### Tests / conventions

15. The marketplace fixtures in `tests/cli_endpoints/conftest.py:34` do not clear `CHAP_MARKETPLACE_URL`, so the tests fail if it is set in the environment. Add `monkeypatch.delenv("CHAP_MARKETPLACE_URL", raising=False)`.
16. The branch name `eas/clim-1085-...` does not follow the `feat/` prefix convention in CLAUDE.md.

ivargr:
@claude review

ivargr (approved):

PR #591, opened 2026-09-11
feat(thresholds): add geometric endemic channel strategy

knutdrand:
## What

Adds a `geometric` endemic channel strategy alongside the existing `seasonal` and `percentile` ones:

```
threshold = expm1(mean(log1p(cases)) + k * std(log1p(cases)))
```

over historical same-month (or same-week) values — the geometric mean times the geometric standard deviation raised to the power `k`. `std_multiplier` takes a scalar or a list, so one request can return several lines.

## Why

Case counts are strongly right-skewed, and `seasonal`'s arithmetic mean is pulled up by single epidemic years. Working on a log scale centres the channel on the geometric mean instead, which AM-GM keeps at or below the arithmetic mean for any non-constant history.

Computing on `log1p`/`expm1` rather than a plain geometric mean is deliberate: it keeps zero counts, which a plain geometric mean cannot represent, and an all-zero season yields a finite threshold of 0 rather than dropping the rows.

## Caveat, stated in the strategy description

Only the *centre* is guaranteed lower than the arithmetic channel. The band is multiplicative, so the log-scale standard deviation is a relative spread, and on a short baseline a single extreme year can make the back-transformed band **wider** than `mean + k*std`:

| baseline (one 20x spike) | `seasonal` | `geometric` |
|---|---|---|
| n=5 | 2179 | **2652** (wider) |
| n=8 | 1681 | 1211 |
| n=20 | 1045 | 445 |

That matters because `percentile` defaults to a 5-year baseline, which is exactly where the effect reverses. The registry description, module docstring and contributor guide all say this, and the test asserts the guaranteed property (`k=0`, AM-GM) rather than the data-dependent inequality.

## Notes

- No endpoint changes — registration through `@threshold` plus membership in the `ThresholdParams` union is enough.
- Contributor guide gains a built-in strategy table covering all three strategies and when to reach for each.
- An earlier revision of this branch also added a `median_iqr` (Tukey fence) strategy; it was dropped as redundant next to `percentile`.

## Testing

- `make lint` clean (ruff, mypy, pyright)
- Full suite: 1501 passed, 122 skipped, 4 xfailed
- New tests cover values, parity with the module-level helper, zero counts, the AM-GM centre against an epidemic-year baseline, multi-line output, weekly data, frequency mismatch, and params validation, plus an end-to-end endpoint test

### DIFF
--- a/chap_core/assessment/thresholds/__init__.py
+++ b/chap_core/assessment/thresholds/__init__.py
@@ -73,6 +73,7 @@ def list_threshold_strategies() -> list[dict]:
 
 def _discover_strategies():
     from chap_core.assessment.thresholds import (
+        geometric,
         percentile,
         seasonal,
     )
--- a/chap_core/assessment/thresholds/geometric.py
+++ b/chap_core/assessment/thresholds/geometric.py
@@ -1,0 +1,87 @@
+"""Geometric threshold strategy: the endemic channel on a log1p scale.
+
+Case counts are strongly right-skewed, so an arithmetic mean is pulled up by single
+epidemic years. This channel takes the mean and standard deviation of ``log1p(cases)``
+instead and back-transforms with ``expm1``, giving a multiplicative channel: each line is
+the geometric mean times the geometric standard deviation raised to the requested power.
+Working on a ``x + 1`` scale keeps zero counts, which a plain geometric mean cannot
+represent.
+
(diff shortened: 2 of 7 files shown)


ivargr:
Deleting this PR, after discussion that we do not want it (instead add stuff when people request)

PR #603, opened 2026-09-14
feat: expose metric unit and target in the metrics API (CLIM-1103)

edvinstava:
Adds optional `unit` and `target` metadata to metric specs and exposes them on the metric catalogue endpoint, so the frontend can drop its hardcoded MAPE unit and coverage/bias targets.

- `unit` is a display suffix for the raw score (`%` for MAPE, `periods` for peak period lag). It does not rescale the score.
- `target` is the ideal raw value for metrics where neither higher nor lower is better (`0.8` for coverage 10-90, `0.5` for coverage 25-75 and ratio above truth, `0.0` for the peak metrics).
- Convention: a metric with `optimization_direction=None` sets a `target` instead, and a score closer to it is better. Documented on the API field and in the contributor guide.
- Score calculations and metric descriptions are unchanged for now.

Companion to [frontend #255](https://github.com/dhis2-chap/chap-frontend/pull/255), which still needs to consume the new fields.


knutdrand:
Coverage metrics are imo (fighing with Claude about this) better the higher up til the target where its flat. That should be the message when showing it in FE. Only too low scores get a bad color. Should be reflected in API docs.

knutdrand:
I think the merge caused an error here with some preexisting work where Claude meant optimizing sensitivity and specificity was bad for HPO, but for info to user I think it should be ok, so reset these to the correct values @edvinstava

knutdrand:
Im 5050 about the validi_hpo_objective flag. Other metrics should maybe set it to False as well (like coverage) and if True by default it might oversell metrics as hop_objectives. Maybe I would wait for the concrete hpo solution to land before implementing it, but not sure. can do a huddle on it

knutdrand (approved):

PR #608, opened 2026-09-15
fix: store chapkit git_revision as model template source_digest

ivargr:
## Problem

Chapkit 2.0 services report the commit they were built from as `git_revision` on `/api/v1/info`. chap-core already parsed the field (#564) but dropped it before it reached `modeltemplatedb`, so every chapkit template was stored with `source_digest = NULL`. That made the sha-based marketplace lookup (CLIM-980) apply to none of the models the marketplace contains. Closes CLIM-1112.

## Change

- Registry sync (`crud.py`): pass `service.info.git_revision` as `source_digest`.
- Startup seeding (`model_template_seed.py`): `ExternalChapkitModelTemplate.get_model_template_config_with_digest()` returns the template config together with the reported commit from a single `/api/v1/info` call; the seed passes the digest through and closes the client when done. `get_model_template_config()` delegates to it, so other callers are unchanged.
- `MLServiceInfo.git_revision` is normalised on parse: only a commit sha (7 to 40 hex, lowercased) is kept. Blank values and placeholders such as `unknown` or an unexpanded `${GITHUB_SHA}` become `None`, so a bad build cannot freeze a garbage digest on the write-once row.

The digest is written only when a row is created. An existing row with `NULL` is left as is: backfilling it would label the backtests that already reference the row with a revision that was never observed when they ran. What happens when a service reports a different sha under an existing version label is decided in CLIM-1113; the existing write-once behaviour (keep the first digest, log a warning) is unchanged here.

`ModelTemplateConfigV2` is unchanged; the sha is threaded alongside it rather than added to the YAML config shape.

## Tests

- Registry path: a service reporting a sha, `None` and `""` stores the sha, `NULL` and `NULL`; a different sha under the same version keeps the first and warns.
- Seed path: chapkit service with and without a reported revision.
- Template: the real body of `get_model_template_config_with_digest()` for a sha, `None` and `""`, with a single info call.
- Parser: whitespace, case, short sha, blank, placeholder and unexpanded-variable inputs.

`make lint` passes. `make test` passes apart from the pre-existing local redis `unknown command HELLO` errors in `test_celery.py`, `test_jobs_routes.py` and `test_db_endpoints.py`, which fail identically on master.


mortenoh:
Reviewed the diff. The change does what it says and CI is green, but two things should be fixed before this is approved. The rest is polish.

## Fix before approving

**1. The NULL-digest fill ignores drift** (`chap_core/database/database.py:177`)

The fill runs after `drifted_template_content_fields` but does not consult its result. A template seeded from an image without `GIT_REVISION` (digest NULL, description "x"), then rebuilt from commit `bbbb` with description "y" under the same version, keeps "x" but gets stamped with `bbbb`. The keep-first rule at 156-166 then defends that wrong pairing forever, and a later sync from the commit that actually produced "x" is rejected. Neither new test exercises drift and fill together. Gate the fill on `not drifted` and log instead.

**2. `get_source_digest()` re-issues the `/api/v1/info` call** (`chap_core/models/external_chapkit_model.py:314`)

`get_model_template_config()` already calls `client.info()` and discards `git_revision`. The new method calls it again, so startup seeding makes a third blocking round trip per chapkit model, and a transient failure on the redundant call now hits the catch-all at `model_template_seed.py:122` and skips seeding the whole model and its configured models. Behind a rolling deploy the two responses can also come from different builds. Return the digest from the info already fetched, the way `crud.py` reuses `service.info.git_revision` without a new request.

## Worth doing in the same PR

**3. Empty string is stored as-is** (`crud.py:142`, `external_chapkit_model.py:314`). `MLServiceInfo.git_revision` has no validator, so a service reporting `""` stores `""`, which the `is None` fill never repairs, and migration `b8c9d0e1f2a3` already treats unknown as NULL. Stock chapkit normalises to None, so this is only reachable from forks or raw registration payloads. `or None` at both sites, plus a test registering `""` and asserting `sourceDigest` is None.

**4. Extra commit** (`database.py:179`). The fill commits directly above the existing archived-reset commit at 183, so a re-sync of an archived, digest-less template does commit -> expire -> reload -> commit -> requery -> commit. Set both fields under their own conditions and commit once at the end of the branch. Do not defer to a downstream commit: `_make_live_template_version` early-returns without committing and the seed path has no follow-up commit.

**5. Comments and docstring now contradict the code.** "A version is write-once, so keep the stored row" (174), "do not change its contents" (180), "Other sources give None" (185), and the column docstring "It cannot change." (`model_templates_and_config_tables.py:137`) are all now false. A simpler shape is one block where the mismatch check is now: `if digest is not None: if existing is None: fill; elif differs: warn`, with one accurate comment.

**6. `get_source_digest()` is never executed by any test.** Both fakes stub it, and the integration tests go through `crud.py`. `tests/external/test_external_chapkit_model.py:168` already has `_make_template_with_schema` with a MagicMock client whose `info()` returns an `MLServiceInfo`; a two-line test there for a set revision and the None default would cover the real body.

## Minor

**7.** `test_seed_stores_chapkit_git_revision_as_source_digest` (`tests/test_database.py:645`) copy-pastes the `FakeChapkitTemplate` from the test above it with one method changed, and the digest-less but valid service is still named `broken_chapkit`. One module-level fake parametrised by digest/version would serve both.

**8.** The conflict warning at `database.py:163` still says "its ref now points to", which is git wording, and now fires for chapkit re-registrations that have no ref. Source-neutral wording such as "now reports revision {new}" fits both paths.

**9.** `test_resync_fills_source_digest_and_keeps_the_first_revision` bundles fill-if-NULL and keep-first-and-warn, and the second half depends on the first half's fill. Both digest behaviours are already unit-tested in `test_database.py:366-388`; line 131 is the only assertion anywhere that the drift warning is emitted, so keep that assertion when splitting.

Items 1 and 2 are both one-line-shaped fixes. Once they land I am happy to approve.

ivargr:
Thanks for the thorough pass. Since you reviewed, the NULL-fill was removed entirely (commit 77e661a8): writing a digest onto an existing row would label the backtests that already reference it with a revision that was never observed when they ran. The digest is now written only when a row is created. What happens on a mismatch under an existing label is being decided in CLIM-1113. That resolves 1 and 4 and the fill-related parts of 5 and 9 by removal. The rest is in the latest commit:

1. Moot: the fill is gone, so drift and digest can no longer be paired wrongly.
2. Done. `get_source_digest()` is removed. `get_model_template_config_with_digest()` makes one `/api/v1/info` call and returns the config with the reported commit; the seed path uses it, and `get_model_template_config()` delegates to it so other callers are unchanged.
3. Done. `or None` at both sites, plus a registration test for `""` asserting `sourceDigest` is null and a unit case for the same in the template test.
4. Moot: no extra commit without the fill.
5. The write-once comments and the column docstring are true again now that nothing writes to an existing row. The "Other sources give None" comment is updated to name git and chapkit as digest sources.
6. Done. `TestGetModelTemplateConfig` now exercises the real body for a set revision, the None default and empty string, and asserts a single info call.
7. Done. One module-level `_fake_chapkit_template` parametrised by version and digest per name, and the config-dir helper takes hostnames, so the digest-less service in the digest test is `unpinned_chapkit`.
8. Done. The warning now says "its source now reports revision".
9. The fill half of that test is gone. The remaining test registers sha A, re-registers sha B under the same version, and keeps the assertion that the warning is emitted.

mortenoh:
## Review

No crash or exception bugs found. The mechanical change is sound: `MLServiceInfo.git_revision` exists on both the registry and wrapper paths, all callers of the renamed method still work, `source_digest` is excluded from content-drift checks, and the new tests pass. The findings below are about behaviour the PR newly triggers, an untested upgrade path, and cleanup.

### Behavioural

1. **Warning fires on every template list and every registration.** `chap_core/database/database.py:160` now logs the "CHAP keeps the first revision" warning whenever a chapkit service is rebuilt under an unchanged version label. Chapkit's scaffold defaults `version` to `1.0.0` and bakes `GIT_REVISION` from HEAD on every build, so in the normal dev loop the stored digest never converges. `_sync_live_chapkit_services` runs unconditionally from `list_model_templates` and `register_service`, so every frontend poll logs one warning per rebuilt service, with no dedup or rate limit. `test_republished_service_under_the_same_version_keeps_the_first_revision` locks this in.

2. **Stored digest can describe code that is not running.** Write-once semantics are borrowed from git templates, but CHAP does not pin chapkit services to the stored sha (`_resolve_chapkit_live_source_url` has no pin). After a rebuild without a version bump, `GET /v1/crud/model-templates` reports the old sha while the live service runs new code. Before this PR the field was honestly null. I see the same-label policy is deferred to CLIM-1113; worth deciding whether shipping a possibly stale sha ahead of that is acceptable.

3. **Existing chapkit rows never get a digest, silently.** Every chapkit row created before this PR has `source_digest` NULL. The mismatch guard at `database.py:156` requires the stored digest to be non-None, so those rows never receive a value (the only write is on the discarded new object) and drift detection is permanently off with no log line. The tests only insert fresh rows, so this upgrade path is untested. Not backfilling is stated as intentional; a one-line `logger.info` on the None-to-value transition would at least surface it without changing write-once.

4. **Only the empty string is normalised.** `git_revision or None` at `chap_core/models/external_chapkit_model.py:314` and `crud.py:142` collapses `""` only. Whitespace-only values or placeholders such as `unknown` or an unexpanded `${GITHUB_SHA}` are stored verbatim and then frozen by the write-once guard, so the real sha reported after the build is fixed is rejected forever. A `.strip()` plus a hex-length check would harden it.

### Cleanup

5. **Comment states an invariant the code does not hold.** The new comment near `database.py:181` says chapkit services give a digest. Both chapkit call sites map a missing revision to None, `schemas.py:105` declares `git_revision` optional for 1.x services, and the PR's own tests exercise the None path. Suggested wording: "Git templates always give a digest; chapkit services give one only when they report git_revision."

6. **Duplicated normalisation.** The empty-to-None mapping is written at two consumers and tested in two places. A `@field_validator("git_revision", mode="before")` on `MLServiceInfo` in `chap_core/rest_api/services/schemas.py` (the file already uses one on `id`) would cover both and let the call sites pass `.git_revision` straight through. Caveat: the unit test builds the `""` case via `model_copy(update=...)`, which skips validation, so it would need `model_validate`.

7. **Unclosed httpx client in the seed path.** `chap_core/database/model_template_seed.py:106` constructs `ExternalChapkitModelTemplate(config.url)`, which opens an `httpx.Client`, and never closes it. The sibling path in `crud.py` does `try/finally client.close()`, and the class already implements `__enter__`/`__exit__`. Pre-existing, but the PR edits this block.

8. **Three near-identical registration tests.** `test_chapkit_self_registration.py:100` onward has three tests with the same body differing only in input; the "without" and "empty" variants also pass on master unchanged. One `@pytest.mark.parametrize("git_revision, expected", [("a"*40, "a"*40), (None, None), ("", None)])` test would mirror the table already used in `tests/external/test_external_chapkit_model.py:192` in this PR.

9. **`call_count == 1` only guards the seed path.** The `name` property still calls `client.info()` on every access, so `get_model()` issues two or three `/info` GETs. Non-blocking; either drop the assertion or memoise the service info, though the latter is beyond the fix.

10. **Test fake indirection.** `_fake_chapkit_template` in `tests/test_database.py:598` derives the name via `url.split("//")[1].split(":")[0].replace("-", "_")` and two name-keyed dicts, so a reader has to trace the dash-to-underscore mapping to see why `versions={"broken_chapkit": None}` affects host `broken-chapkit`. Taste.

### PR description

The body references `ExternalChapkitModelTemplate.get_source_digest()`, which no longer exists after the rename to `get_model_template_config_with_digest()`. It also opens with Summary rather than a Problem section.

ivargr:
Second round. As agreed offline, 1 to 3 are the same-label policy and the upgrade path, which CLIM-1113 covers, so they are not addressed here.

4. Done, and folded into 6: a `mode="before"` validator on `MLServiceInfo.git_revision` strips, lowercases and keeps only a 7 to 40 hex sha (`COMMIT_SHA_PATTERN` from `chap_core/external/github.py`); anything else becomes `None`. The two call sites now pass `.git_revision` straight through. Parser test covers whitespace, case, short sha, blank, `unknown`, `${GITHUB_SHA}` and `None`.
5. Done, your wording.
6. Done, see 4. The template unit test builds the info with `model_validate` so the validator runs.
7. Done. The seed path uses the template as a context manager so the client is closed after the info and schema are fetched.
8. Done, one parametrised test with the same table as the template test.
9. Kept the `call_count == 1` assertion. It only claims what the method does, and it is the regression guard for your point 2 from the first round. Memoising info across the class is out of scope here.
10. Done. The fake's name is now `urlparse(url).hostname`, so the dicts are keyed by the host as written in the YAML with no mapping to trace.

PR body updated: it opens with a Problem section and names the current method.

mortenoh (approved):

PR #610, opened 2026-09-16
feat(logs): expose complete job logs and chapkit model output

knutdrand:
## Problem

Running the chapkit version of ewars through the modeling app, a failed backtest returned only this from `GET /v1/jobs/{id}/logs`:

```
Starting backtest for model 'chapkit-ewars-model' on dataset ID 12
Validating dataset with 18 locations
Running 10 evaluation splits with prediction length 12
```

plus the celery traceback, whose innermost message was chapkit's truncated stderr tail. Two separate things were dropping output.

**1. The logs endpoint served a filtered subset.** `TrackedTask` wrote two files per task: `task_<id>.status.txt` (the `chap_status` logger, ~10 call sites in `db_worker_functions.py`) and `task_<id>.debug.txt` (this module's logger plus the root logger, so everything from chap-core and third-party packages). `CeleryJob.get_logs` read only the status file.

**2. Nothing fetched the model's stdout/stderr.** The R script runs inside the chapkit service, not the chap-core worker. On failure, chap-core used only `job.error` and `job.error_traceback`, and chapkit inlines a truncated stderr tail there. The complete output sits on the run's diagnostic artifact, which was never read.

## Changes

- `CHAPKitRestAPIWrapper.get_run_output` reads `metadata.stdout` / `metadata.stderr` off an ML artifact, validated with chapkit's `validate_artifact_data`. Best-effort with a short `RUN_OUTPUT_TIMEOUT`: an unreachable or unparseable artifact logs a warning and returns `""`, so a diagnostic lookup can never mask or stall the failure it is describing. Each stream is cut to its last `MAX_RUN_OUTPUT_CHARS` (64k) because the exception message is persisted in the job metadata and re-sent on every logs poll.
- `_failure_message` in `external_chapkit_model.py` appends that output to the `ModelFailedException` raised by `train` and `predict`, reading the artifact id pre-allocated by the `$train` / `$predict` response. chapkit 2.0.1 stores the diagnostic artifact under that id and never sets `job.artifact_id` on the failure path. A cancelled job leaves no artifact, so nothing is fetched.
- `CeleryJob.get_logs` serves the last `LOG_TAIL_BYTES` (256k) of the per-task log, cut to a line boundary and marked when truncated, falling back to the traceback with a warning when the file is missing.
- `TrackedTask` points the status logger at the per-task log handler so progress messages are interleaved chronologically with the rest of the output. The `task_<id>.status.txt` file is removed; nothing read it any more. The comments, `get_status_logger` docstring and docs that described the old status/debug split are updated.
- `alembic/env.py` called `fileConfig` with the default `disable_existing_loggers=True`, which disabled the already-created `chap_status` logger in any process that ran migrations after importing the worker functions, silently dropping progress messages. It now passes `disable_existing_loggers=False`, and `get_status_logger` re-enables the logger so a later logging config call cannot silence it.

A failed run now reports:

```
Prediction job 01M2... ended with status 'failed': <chapkit message>. Stacktrace: ...
Model output:
--- stdout ---
...
--- stderr ---
<the INLA output, up to the last 64k characters>
```

## Tests

- `tests/external/test_chapkit_integration.py::TestFailedRunOutput` - full output surfaced on a failed train, the predict path fetches the pre-allocated prediction artifact, failure still raised when the artifact is unreadable or does not validate, empty result when metadata carries no streams, per-stream truncation.
- `tests/test_celery.py` - the per-task log is served, a large log is served as a bounded tail, and an eager `TrackedTask` run writes both status-logger and library output to the per-task log.
- `tests/test_log_config.py` - `get_status_logger` re-enables a logger that a logging config call disabled.
- `test_predict_raises_on_cancelled_job_without_fetching_artifact` keeps asserting that a cancelled job triggers no artifact fetch.

`make lint` and `make test` clean.

## Note for review

The served log is the root-logger stream, so it now includes httpx request lines, the config payload logged by the chapkit model and, under `CHAP_DEBUG`, echoed SQL. Auth on the jobs routes is opt-in via `CHAP_API_TOKEN`. This PR does not add a gate; whether one is wanted is a separate decision.


mortenoh:
## Review

Ten findings, most severe first. The first three are about what the logs endpoint now exposes and how much data it moves, and are worth settling before merge.

### Security and scale

1. **`GET /v1/jobs/{id}/logs` now serves the root-logger debug file with no filter or gate.** Auth is opt-in (`chap_core/rest_api/auth.py:96-98` bypasses when `CHAP_API_TOKEN` is unset) and `jobs.py:150-166` has no per-route dependency. `TrackedTask` attaches the debug handler to the root logger (`celery_tasks.py:124`) and workers run at `--loglevel=info`, so the served file contains httpx `HTTP Request: POST http://<internal chapkit host>/...` lines, the full `config_data` payload logged at `external_chapkit_model.py:255`, and under `CHAP_DEBUG` every SQL statement with bound values (`database.py:53` uses `echo=is_debug_mode()`). Meanwhile `celery_tasks.py:105` still says "(full debug logs, server access only)", `log_config.py:75-83` still says the status logger exists "to prevent leaking sensitive debug information", and `docs/modeling-app/troubleshooting/logs.md:33-34` and `docs/contributor/rest_api_and_database.md:99` still document the old split. Either make this a deliberate, gated change (config flag or auth requirement) or at minimum update the three stale contracts in this PR.

2. **Full, uncapped model stdout+stderr is embedded in the `ModelFailedException` message and persisted in Redis with no expiry.** chapkit does not truncate artifact stdout/stderr; only `job.error` gets `format_stderr_tail`, which this PR deliberately bypasses. `get_run_output` joins both streams verbatim and `_failure_message` appends them. Nothing catches `ModelFailedException`, so it reaches `TrackedTask.on_failure` (`celery_tasks.py:215-229`), which hsets both `error=str(exc)` and `traceback=str(einfo.traceback)` (the traceback ends with the message) into `job_meta:{id}`, keys that never expire; the Celery result backend stores a third copy. `get_logs` (`celery_tasks.py:354-356`) then appends the traceback to every FAILURE `/logs` response. A verbose R/INLA model printing per-iteration progress for an hour yields multi-MB Redis entries per failed job and multi-MB `/logs` payloads. Cheaper: log the output once at the failure point with `logger.error` (it lands in the served debug file, in chronological position) and keep the exception message short, or cap each stream (for example the last 64 KB) before it enters the message.

3. **`get_logs` reads the entire, ever-growing debug file and returns it whole on every poll.** The route docstring (`jobs.py:154-166`) still says "Tail the log output", but `job.get_logs()` is called with no arguments and `n_lines` exists only on `NaiveJob`. httpx logs `HTTP Request: ...` at INFO per request and `wait_for_job` logs `Job {id} status: ...` every 2s (`chapkit_rest_api_wrapper.py:278`), so a multi-hour backtest with 10 splits produces a multi-MB file that is fully re-read, JSON-encoded and shipped to the frontend on each progress poll. Previously the endpoint served the small status file. Seek to `max(0, size - N)` and return a tail, or accept an offset/`n_lines` query param.

### Correctness and dead code

4. **The diagnostic artifact GET is issued on every non-completed status including `canceled`, and uses the 300s read timeout.** chapkit writes no artifact on cancel (`runner.py:707-710` rmtree's the workspace and re-raises before `artifact_manager.save` is reached), so that fetch is a certain 404 plus a warning. `get_run_output` passes no timeout override, so `DEFAULT_REQUEST_TIMEOUT` (connect=10, read=300) applies: a chapkit container that reports the job failed and then hangs blocks `train()`/`predict()` for ~310s before raising. Skip the fetch when `job.status == "canceled"` and pass a short explicit timeout for this best-effort request.

5. **The `job.artifact_id` candidate, `getattr` guard, `dict.fromkeys` dedup and loop at `external_chapkit_model.py:322` are dead.** `uv.lock` and `pyproject.toml:61` pin chapkit 2.0.1, not 1.1.0 as the PR description says. In 2.0.1, `manager.py:469-474` notes that the scheduler never calls `_on_job_result` on the failure path, so the artifact id is carried in the error message rather than `ChapkitJobRecord.artifact_id`, and the diagnostic artifact is saved under the pre-allocated id before `ModelRunFailedError` is raised (same for predict at `:578-609`). So `getattr(job, "artifact_id", None)` is always None, the loop iterates once, and the "404 falls back to today's truncated form" caveat cannot occur. Simpler: `_failure_message(kind, job, artifact_id, client)` with one `client.get_run_output(artifact_id)` call, and correct the PR note. If a fallback is wanted, the sanctioned source is the `diagnostic artifact <ulid>` text in `job.error`.

6. **`task_<id>.status.txt` is now dead output and the `get_logs` fallback is unreachable.** Repo-wide the only references to `.status.txt` are the writer (`celery_tasks.py:111`) and the fallback reader (`:348-349`). `logging.FileHandler` defaults to `delay=False`, so both files are created at construction before `super().__call__` runs, and master already created the debug handler first, so no worker version ever produced a status file without a debug file. `test_get_logs_falls_back_to_status_log` locks in a path production cannot hit. Since the API now serves the debug file, the smaller change is `status_logger.handlers = [debug_file_handler]`, then delete the status handler/file, the fallback branch and its test, and fix the now-false comment at `celery_tasks.py:105` plus the docs.

### Tests

7. **`artifact_calls` in `tests/external/test_chapkit_integration.py:657` is dead plumbing** after `assert artifact_calls == []` was removed. The mock 404s every `/api/v1/artifacts/*` request and `get_run_output` swallows that and returns `""`, so `"Model output" not in str(exc_info.value)` holds whether or not any fetch happens or which id it targets. Assert `artifact_calls == [f"/api/v1/artifacts/{VALID_ULID_2}"]` (the id pre-allocated by the mock predict response at `:630`) to make the test meaningful, or drop the list and the handler parameter.

8. **The `TrackedTask` handler-wiring change has no test.** Reverting `celery_tasks.py:136` to `[status_file_handler]` leaves the suite green while `/logs` loses every progress line; the only end-to-end test, `test_celery_logging`, is skipped as "Not stable". `tests/test_celery.py:110-124` already monkeypatches `CHAP_LOGS_DIR` and `r` with `task_always_eager=True`, and Celery's eager tracer dispatches to the custom `__call__`, so queuing a function that calls `get_status_logger().info("progress")` and asserting on `tmp_path/task_<id>.debug.txt` is a few lines. Separately, `TestFailedRunOutput` only covers `model.train`; `_failure_message("Prediction", ...)` with a readable artifact (`external_chapkit_model.py:384-387`) is never exercised.

### Minor

9. **`get_run_output` (`chapkit_rest_api_wrapper.py:194`) hand-walks `artifact.data` with isinstance/`.get` guards** instead of validating with chapkit's `MLMetadata` schema like `get_prediction_artifact_dataframe` five lines above. A renamed or re-nested metadata key returns `""` with no log line distinguishing "no output recorded" from "shape changed". Moving `validate_artifact_data(artifact.data)` inside the existing `try` surfaces drift as a logged `ValidationError` and gives typed access plus `exit_code`. Trade-off: `BaseArtifactData` has `extra="forbid"`, so additive chapkit fields would also trip it.

10. **Per-task log capture swaps handlers on process-global loggers** (`celery_tasks.py:118-137`). Safe under the default prefork pool used by every shipped launch, but `--pool=threads`/gevent or eager execution in a shared process would put job B's root-level records into `task_A.debug.txt`. This cross-talk pre-existed for the one-line status file; the PR widens it to the whole root stream. A one-line comment stating the prefork assumption is enough.

### Note

In the Helm chart the api and worker pods use separate emptyDir volumes for `/data/logs` (`charts/chap/templates/{api,worker}/deployment.yaml`), so the logs endpoint always hits the traceback fallback there. Pre-existing, but the removed `logger.info("Looking for log file at ...")` was the only breadcrumb; a `logger.warning` on the fallback branch would be a better replacement.

ivargr (approved):

PR #616, opened 2026-09-18
feat!: refuse a changed model revision under an existing version label (CLIM-1113)

ivargr:
Implements CLIM-1113.

A model template is identified by `(name, version)` and its `source_digest` is write-once. Until now a changed revision under an existing label was silently ignored: git templates kept running the old code while the operator believed they had deployed new code, and chapkit templates ran new code under a stored digest that no longer matched. This PR makes the digest guard the label on both paths.

## Git path (startup)

- Every entry under `versions:` in `config/configured_models/*.yaml` must be a full 40-character commit sha (with or without `@`). Branches, tags and short shas fail at parse time, before any network or database access, with a message naming the file, the label and the value. Every label is checked, not only the last one.
- Reseeding a label whose sha differs from the stored digest raises `ModelTemplateRevisionConflict` and fails startup, naming the label, the stored sha and the new sha. A stored row without a digest (seeded before digests existed) keeps its row.
- `default.yaml`: the ten branch and tag entries are removed. Every seeded (last) entry was already a full sha, so no stored row changes.

## Chapkit path (read time and run time)

- `_sync_live_chapkit_services` returns the revision conflict per registered template name (None when live) instead of a set of live ids. A mismatched template is left untouched and its configured models are not re-synced. The mismatch is computed on every read, not persisted, so redeploying the stored revision clears it without cleanup.
- A service that reports no revision is not stored at all, so its version label stays free for the build that does report one. The registration response and chap's log say to build with the `GIT_REVISION` build arg.
- Config-seeded chapkit templates (not in the registry) log the same conflict at seed time, since a refused job would otherwise be the first signal.
- `ModelTemplateRead.health_status` is typed `Literal["live", "revision_mismatch"] | None`, so the OpenAPI schema carries the enum.
- `GET /v1/crud/model-templates` reports `healthStatus = "revision_mismatch"`; the field description on `ModelTemplateRead` documents the value for the modeling app.
- `POST /v2/services/$register` still succeeds on a mismatch but appends the conflict text to the response message so the model developer sees it in their service logs.
- `get_configured_model_with_code` compares the revision the service reports now with the stored digest before running, so backtests and predictions in the celery worker fail with the conflict message instead of producing a mislabelled result. A missing revision on either side is a mismatch.
- The write-once digest warning in `add_or_update_model_template` is removed. The method never writes a digest to an existing row.

## BREAKING CHANGE

Existing deployments stop running their models until the operator or model developer acts. Refusing is deliberate: a model whose revision cannot be matched to its stored version must not produce results under that version's label.

Every chapkit model must bump its `info.version` once after this ships: all chapkit templates stored today have a NULL digest, which now counts as a mismatch. Chapkit images built without the `GIT_REVISION` build arg report no revision and are refused to run (CLIM-1111 fixes the build files). Operators with a local `config/configured_models/*.yaml` that uses a branch ref must pin it to a commit sha before upgrading, or CHAP will not start.

`compose.ewars.yml` pins the ewars image to the build that reports `info.version` 1.0.1, which is the version bump this release note asks every chapkit model to make.

## Deviations from the issue notes

- A git row stored without a digest (seeded before PR #527) is kept rather than treated as a conflict, so legacy deployments keep starting.
- `add_model_template_from_url` re-checks by `(name, version)` after fetching the MLProject. The by-URL lookup misses when the stored source URL differs, and `add_or_update_model_template` would then silently reuse the row.


mortenoh:
Review of aba65154. The design holds up and the tests cover the mismatch matrix on both the git and chapkit paths. I would fix finding 1 before merge, because it makes a version label permanently unusable. The rest are discussion points and nits. This is from reading the diff; I did not run anything locally.

## Findings, most severe first

**1. A chapkit version registered once without `GIT_REVISION` can never run (medium)** -- `crud.py`, `_sync_live_chapkit_services`

- When no row exists for `(id, version)`, the sync inserts one with `source_digest=None` and only then computes the conflict. Rows are write-once, so the label stays NULL.
- Rebuilding the same version with the build arg is then a permanent mismatch, stored `None` against a reported sha. The parametrized case `(None, "a" * 40)` asserts this as intended behaviour, so the developer has to bump the version a second time.
- The NULL row never gets configured models, so `_make_live_template_version` never makes it live. The list endpoint keeps showing the old row with `revision_mismatch`, while the logged conflict text names the new version with stored `None`. That is confusing to debug.
- Suggested fix: when `service.info.git_revision is None` and no row exists, report the conflict without persisting the row. `test_service_without_a_git_revision_gets_no_configured_models` would need a matching update.

**2. Legacy git rows are never guarded (medium, may be deliberate)** -- `model_template_seed.py`, `add_model_template_from_url`

- A row with a NULL digest is kept, and `add_or_update_model_template` never backfills the digest.
- For deployments seeded before #527, an operator can still change the sha under an existing label silently, which is the case this PR sets out to prevent.
- One option: when the stored `source_url` already ends in a full 40-character sha, compare against that, or backfill the digest from it. I did not check what legacy `source_url` values actually contain.
- The asymmetry deserves a line in the PR text: a NULL digest means "keep serving" for git rows but "refuse" for chapkit rows.

**3. Config-seeded chapkit templates get no early warning (low-medium)** -- `uses_chapkit: true` entries in the yaml

- The old digest warning is removed and nothing replaces it at seed time.
- Seeded services that do not also self-register are not in the registry, so `healthStatus` stays `None`.
- The first sign of a mismatch is a failed celery job. Every existing row has a NULL digest, so this hits every such deployment on upgrade.
- A `logger.warning(chapkit_revision_conflict(...))` at seed time would be a cheap fix.

**4. HTTP client leaks on the refusal path (low)** -- `database.py`, `get_configured_model_with_code`

- `ExternalChapkitModelTemplate(source_url)` opens a `CHAPKitRestAPIWrapper` in its constructor. Raising the conflict never closes it.
- Closing the client before `raise conflict` would fix it.

**5. Leftover docs and config (low)**

- `docs/modeling-app/managing-model-templates.md` lines 69-77 still show `nightly: "@main"` and say "commits or branches". That section documents the approved-repos whitelist, a different system, but it now contradicts the rule stated a few lines above it.
- `config/configured_models/example2.yaml.disabled` uses tags and branches throughout. Enabling it would now stop startup.
- The PR body contradicts itself: the first chapkit bullet says the sync returns a `live` / `revision_mismatch` status, and a later bullet says it returns the conflict. The first bullet is stale after the third commit.

**6. Nits**

- `health_status` could be typed `Literal["live", "revision_mismatch"] | None`, so the OpenAPI schema carries the enum for the modeling app instead of prose in a description.
- `_registered_chapkit_revision_conflict` matches on `(name, version)` without filtering on `uses_chapkit`, so a git template that shares a service's id and version would be compared too. Unlikely edge case.

## Checks that passed

- Sha case: `FULL_COMMIT_SHA_PATTERN` accepts uppercase hex, and both `resolve_commit_sha` and `MLServiceInfo.normalize_git_revision` lowercase the value, so digest comparisons are consistent.
- Startup does fail: `seed_configured_models_from_config_dir` is called without a surrounding try/except at startup, so both the parse error and `ModelTemplateRevisionConflict` stop it.
- `add_model_template_from_url` has no other callers, and `default.yaml` is the only active yaml with `versions:`.
- Running an older chapkit version's configured model after a version bump is now refused. Before this change it would have run the new code under the old label.

## CI

- `test (windows-latest)` fails on `test_available_metrics_expose_optimization_direction` (`sensitivity.optimization_direction` is `None`). This PR does not cause it: the `master` push runs at 07:49 and 11:21 UTC today fail the same workflow, and it reached this branch through the merge-from-master commit. The run on this branch before the merge was green.
- The ubuntu and macOS test jobs were cancelled as a result, so there is no full test run on the current head yet.

ivargr:
Review points addressed in aed68896:

1. **Burnt label**: fixed. A service that reports no revision is not stored, so the label stays free for the build that reports one. The registration response and chap's log say to build with `GIT_REVISION`. The config-dir seed path skips such a service for the same reason.
2. **Legacy git rows with NULL digest**: left as is. Nobody is in production yet, and the deviations section of the PR body states the asymmetry.
3. **Seed-time warning**: added. Config-seeded chapkit templates log the conflict when the service reports another revision than the stored version.
4. **Client leak**: fixed. `ExternalChapkitModelTemplate.close()` is called before the conflict is raised.
5. **Docs and config**: the whitelist example and wording use full shas, `example2.yaml.disabled` uses only full shas, and the stale PR body bullet is rewritten.
6. **Typing**: `health_status` is `Literal["live", "revision_mismatch"] | None`.

Tests added or updated for 1, 3 and 4. The remaining CI failure (`test_available_metrics_expose_optimization_direction`) comes from master through the merge commit and is not touched by this branch.

mortenoh (approved):

ivargr:
eb0aedbc fixes the two slow-test failures in `tests/integration/test_chapkit_e2e.py`. They were caused by this PR: the fixture service reported no git revision, so the run-time check refused the seeded template. The fixture now sets `GIT_REVISION` and the template is stored from the revision the service reports, which is what the real registration path does. Verified locally with `--run-slow`: 5 passed.

